PR #79, opened 2017-11-23
Adding tests - help command + Angular add template command

dafo:


damyanpetev:
Talking with @dafo we generally agreed that we have in the current acceptance folder is a mix of some doing what is essentially unit testing. So we do want to split those in unit and e2e(acceptance-like, but popular enough term).
I'll merge these and we'll apply the changes separately.

PR #81, opened 2017-11-24
feat: add editors templates for Angular

### LINKED ISSUE #16: Implement Editors templates
o	Text editor
o	Numeric
o	Currency Editor
o	Mask Editor
o	DatePicker

Lipata:
closes #16


templates/angular/ts/editors/index.ts
```diff
@@ -0,0 +1,38 @@
...
+	["currency-editor",  {name: "Currency Editor", widget: "ig-currency-editor"}],
+	["mask-editor",  {name: "Mask Editor", widget: "ig-mask-editor"}],
+	["date-picker",  {name: "Date Picker", widget: "ig-date-picker"}]
+]);
+
+class EditorsComponent extends MultiTemplateComponent<EditorsAngularComponent> {
+	constructor() {
+		super(EditorsAngularComponent, templates);
+		this.name = "Editors";
+		this.group = "Editors";
```
damyanpetev: should be "Data Entry" to match other frameworks.

templates/angular/ts/editors/default/files/src/app/components/__path__/__name__.component.ts
```diff
@@ -0,0 +1,17 @@
+import { Component } from '@angular/core';
+
+@Component({
+	selector: "editors",
+	template: `
+		<$(widget) [(options)]="options" widgetId='$(widget)'></$(widget)>
+	`
+})
+export class $(ClassName)Component {
+	public options: Ig$(ClassName);
```
damyanpetev: Note that class name used like this will only work when the default name is selected for the template and gen wrong interface on most user input. A separate template variable based on the control name would be required

PR #105, opened 2017-12-13
IGanchev/step by

### LINKED ISSUE #63: Step by step entry points
Besides the `igniteui` command without arguments, the same should be possible via:
- `igniteui new`: Also starts the guide from project creation
- `igniteui add`: Starts directly at the `Add component/template` step

Question: Should the add version also loop like it would after creating project to allow adding multiple views?

IvayloG:
closes #63

### DIFF
--- a/lib/PromptSession.ts
+++ b/lib/PromptSession.ts
@@ -21,6 +21,7 @@ export class PromptSession {
 
 		add.templateManager = this.templateManager;
 
+		// tslint:disable:object-literal-sort-keys
 		if (config != null && !config.project.isShowcase) {
 			projLibrary = this.templateManager.getProjectLibrary(config.project.framework, config.project.projectType);
 			await this.chooseActionLoop(projLibrary, config.project.theme);
--- a/lib/PromptSession.ts
+++ b/lib/PromptSession.ts
@@ -25,64 +25,64 @@ export class PromptSession {
 			projLibrary = this.templateManager.getProjectLibrary(config.project.framework, config.project.projectType);
 			await this.chooseActionLoop(projLibrary, config.project.theme);
 		} else {
-		//TODO update to check if project exists and load correct framework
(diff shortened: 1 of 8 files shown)


lib/PromptSession.ts
```diff
@@ -21,71 +21,68 @@ export class PromptSession {
 
 		add.templateManager = this.templateManager;
```
bazal4o: // tslint:disable:object-literal-sort-keys should be added back in order tslint to pass

lib/PromptSession.ts
```diff
@@ -22,7 +22,6 @@ export class PromptSession {
 
 		// tslint:disable:object-literal-sort-keys
 		if (config != null && !config.project.isShowcase) {
-			//throw new Error("Add command is supported only on existing project created with igntie-ui-cli");
 			projLibrary = this.templateManager.getProjectLibrary(config.project.framework, config.project.projectType);
```
damyanpetev: So no change here. Not sure this functionality should remain, though we can iron this out later on.

lib/commands/add.ts
```diff
@@ -47,7 +48,13 @@ command = {
...
 		) as ProjectLibrary;
 
-		if (!frameworkLibrary.hasTemplate(argv.template)) {
+		if (!argv.template && !argv.name) {
+			const prompts = new PromptSession(command.templateManager);
+			await prompts.chooseActionLoop(frameworkLibrary, config.project.theme);
+			return;
+		}
+
+		if (argv.template && argv.name && (!frameworkLibrary.hasTemplate(argv.template))) {
```
damyanpetev: The `argv.template && argv.name` check here should be either separate (to show help) or not here at all.
What happens if just name is passed? No feedback at all?
The following `hasTemplate` check kind of does the same too?

PR #120, opened 2018-01-15
Checks for existing files on new/add

### LINKED ISSUE #76: Adding a template with the name of an existing one is overwriting the earlier
**Steps to reproduce:**
1. Create a new project 
2. Add a template 
`add combo myCombo`
3. Add another template using the same command and name

**Actual result:**
The previously created template is replaced by the second template with the same name

**Expected result:**
We should show a message to the user that a template with the same name already exists 
**or**
We should add the second template with incremented name - for example, "myCombo1".

damyanpetev:
Closes #76


dafo:
When creating a project, we should check if the name starts with symbols that are not allowed. For example, "../newProject" should not be allowed for a name.

damyanpetev:
@dafo Seems like we need more checks, currently thinking:
- no slashes or alphanumeric only
- trim names and remove multiple spaces
- react/angular view names will require starting with a letter as those turn into classes.

dafo:
@damyanpetev  I think react view names can start with a number except for 0. 0 is the only restricted number I believe. The angular view names cannot begin with any number.

damyanpetev:
@dafo @bazal4o Should be good to go now.

PR #157, opened 2018-02-06
Implement igx-ts empty project mockup

### LINKED ISSUE #156: Add/Update angular new project tempalate
The default project template should be App Host - NavBar and NavDrawer.

Mockups have been discussed with UX team.

IvayloG:
Closes #156 

•	The content div should be styled so it's width takes the whole available space to the right.
•	Style the Views and items in the NavDrawer (paddings).
•	Links in home component should be styled as left/centered/right.
•	The NavDrawer and NavBar components are using "igniteui-angular": "^5.1.3", and not the latest refactored version of the product. 

@sbozalieva 
* Should the gif remain with the slightly beige it currently uses?
* Are the purple colors used OK, as they are color picked from the mock and seem a bit different (lighter) for the text? Should all the text use the exact color from the NavBar?


IvayloG:
Currently, what is left is to switch from css to scss, as the IgniteUI for Angular components will require it after the new feature is merged.

PR #171, opened 2018-02-14
Automatically commit new project

### LINKED ISSUE #169: Add git options to ig new command
Add options for git init and git commit commands. We should decide that when we create a new project with ig new command:
 - to directly init and commit it, without any defined flags (like in angular-cli), and in the same time have a flags that allow to skip them. In angular-cli the [flags are](https://github.com/angular/angular-cli/wiki/new) --skip-git and --skip-commit.
 - to define two flags that will allow git init and git commit.

We should decide how those flags (and their alias-es) will be named, if commit flag will do both, and if the commit message could be modified.

Lipata:
Automatically commit new project and define options to skip init and commit:

  --skip-git, --sg     Do not automatically initialize repository for the project with Git [boolean] [default: false]

Closes #169


lib/commands/new.ts
```diff
@@ -90,6 +96,22 @@ command = {
 		// TODO: update output path based on where the CLI is called
 		await projTemplate.generateFiles(process.cwd(), argv.name, theme);
 		Util.log("Project Created");
+
+		this.git(argv["skip-git"], argv.name);
+	},
+	git(skipGit, projectName) {
+		if (!skipGit) {
+			process.chdir(projectName);
```
damyanpetev: Can this be passed directly to [exec `cwd` option](https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options)? Will also skip going back afterwards.

spec/acceptance/new-spec.ts
```diff
@@ -52,4 +53,25 @@ describe("New command", () => {
...
 		done();
 	});
+
+	it("Git Status", async done => {
+		const projectName = "angularProj";
+		await cli.run(["new", projectName, "--framework=angular", "--type=igx-ts"]);
+
+		process.chdir(projectName);
+		expect(fs.existsSync(".git")).toBeTruthy();
+		expect(Util.exec("git log -1 --pretty=format:'%s'").toString())
```
damyanpetev: 👍

spec/unit/new-spec.ts
```diff
@@ -1,13 +1,31 @@
+import * as fs from "fs-extra";
+import { parse } from "path";
 import { default as newCmd } from "../../lib/commands/new";
 import { ProjectConfig } from "../../lib/ProjectConfig";
 import { PromptSession } from "../../lib/PromptSession";
 import { Util } from "../../lib/Util";
 import { resetSpy } from "../helpers/utils";
 
 describe("Unit - New command", () => {
+	let testFolder = parse(__filename).name;
```
damyanpetev: Ummm are you sure you need actual folders? Unit tests don't have that setup for a reason - they don't actually write any files at all. There's already a spy on `generateFiles` as far as I see.

Lipata: It's not needed. This was written using the CPDD.

damyanpetev: 😆

PR #177, opened 2018-02-20
#175 Changing the port

### LINKED ISSUE #175: Ignite UI CLI Readme.md update
Currently the instructions of the CLI is is written that the project is generated at localhost:3000 but this is not always the case:

"Upon creation, projects will be automatically served on port 3000 and can be accessed via http://localhost:3000/."

Please update the docs, saying in the end of the project the CLI specify on which url you can find your project.

kdinev:
Closes #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ To get a guided experience through the available options, simply run:
 ig
 ```
 
-Upon creation, projects will be automatically served on port 4200 and can be accessed via http://localhost:4200/.
+Upon creation, projects will be automatically served on port 4200 for Angular project and can be accessed via http://localhost:4200/. **NOTE**: Port may vary for different project types. `ig start` will indicate what port to open the application under in the console.
 
 ### List the available commands.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ To get a guided experience through the available options, simply run:
 ig
 ```
 
-Upon creation, projects will be automatically served on port 3000 and can be accessed via http://localhost:3000/.
+Upon creation, projects will be automatically served on port 4200 and can be accessed via http://localhost:4200/.
```
bazal4o: The port can vary depending on the project type. We can use the paragraph from the wiki page. 
For jQuery projects, navigate to http://localhost:3000/ . The app will automatically reload if you change any of the source files.
Note: For Angular projects you need to navigate to http://localhost:3001/ and for React - http://localhost:3002/

And add the configuration for port 4200 which is the default for ignite ui angular project type.

kdinev: Yeah, that sounds better.

PR #179, opened 2018-02-21
Unit tests for BaseProjectLibrary

dafo:
Adding unit tests for the BaseProjectLibrary


spec/unit/BaseProjectLibrary-spec.ts
```diff
@@ -0,0 +1,324 @@
...
+
+	it("has correct projects.", async done => {
+		const mockProjects = ["angular", "jquery"];
+
+		spyOn(Util, "getDirectoryNames").and.returnValue(mockProjects);
+
+		const library = new BaseProjectLibrary("../test");
+		expect(library.hasProject("angular")).toBe(true);
+		expect(library.projects).toEqual(mockProjects);
+		expect(Util.getDirectoryNames).toHaveBeenCalledWith(path.join("../test", "projects"));
```
damyanpetev: Actually looking at this, seems like the `projects`/ `hasProject` are completely redundant (nothing uses them) and the projects are really not under the root folder :C 
Do we drop this one or make it work?

bazal4o: Currently we have only empty projects. Once we include the option into the new command to choose what type of project  we create (currently type is used to specify the project library for the framework - which is confusing) - we'll need such method. Either comment this out and we'll uncomment it once we enhance the new command, or leave it as it is.

PR #203, opened 2018-02-28
feat: list command is working now

### LINKED ISSUE #68: Add a command to list available templates
Since the number of templates has grow quite a bit, it won't be reasonable to list them in readme/docs.
It'll be better if there's a command that can list all available templates, filtered by framework [and project].

wnvko:
Closes #68


lib/commands/list.ts
```diff
@@ -0,0 +1,103 @@
...
+			Util.log(`'${group}' group:`);
+			for (const template of templatesByGroup[group]) {
+				const spacesCount = maxIdLength - template.id.length + addSpacesCount;
+				const output = "\t" + template.id + spaceChar.repeat(spacesCount) + template.description;
+				Util.log(output);
+			}
+		}
+
+		if (inProject) {
+			Util.log("To list all available templates run list command outside of a project folder");
```
damyanpetev: This should say something in the lines "To list available templates for other framework and project type run outside of a project folder"?

lib/commands/list.ts
```diff
@@ -0,0 +1,103 @@
...
+
+let command: {
+	[name: string]: any,
+	templateManager: TemplateManager,
+	execute: (argv: any) => Promise<void>
+};
+
+// tslint:disable:object-literal-sort-keys
+command = {
+	command: "list [framework] [type]",
```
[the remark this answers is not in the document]
damyanpetev: Also maybe default to "jquery"

lib/commands/list.ts
```diff
@@ -0,0 +1,103 @@
...
+			if (!projectLib) {
+				return Util.error(`Project type '${argv.type}' not found in framework '${argv.framework}'`, "red");
+			}
+		} else {
+			projectLib = command.templateManager.getProjectLibrary(argv.framework) as ProjectLibrary;
+		}
+
+		let maxIdLength = 0;
+		projectLib.templates
+			.filter(t => t.controlGroup !== "Custom Templates")
```
damyanpetev: That "Custom Templates" seems like more of a error in the Angular Editors form template (ig-ts).

PR #207, opened 2018-03-02
fix: change the regex to allow single letter name

### LINKED ISSUE #173: Allow names with just one letter and error fix message
- [x] 1. Allow project and template names with just one letter

**Current State**

> ? Name your component: a
> Name 'a' is not valid. Template names should start with a letter and can also contain numbers, dashes and spaces.> 

**Expected Result**
To create a component named with just one letter.

- [x] 2. Remove template from the message to avoid confusion and fit all scenarios (project/view/component etc) .

**Current State**

> ? Name your **component**: 1
> Name '1' is not valid. **Template** names should start with a letter and can also contain numbers, dashes and spaces.

**Expected Result**

> ? Name your component: 1
> Name '1' is not valid. Names should start with a letter and can also contain numbers, dashes and spaces.

wnvko:
Closes #173

cli.ts - reordering the commands to show correctly in the help
for all commands change the description as discussed with UX
fix all failing tests

### DIFF
--- a/lib/commands/generate.ts
+++ b/lib/commands/generate.ts
@@ -7,7 +7,7 @@ import { default as config } from "./config";
 const command = {
 	aliases: ["g"],
 	command: "generate",
-	desc: "Generates new project",
+	desc: "generates custom template",
 	templateManager: null,
 	// tslint:disable-next-line:object-literal-sort-keys
 	builder: yargs => {
@@ -15,7 +15,7 @@ const command = {
 			.command({
 				aliases: ["t"],
 				command: "template [name]",
-				desc: "Generates custom template",
+				desc: "generates custom template",
 				// tslint:disable-next-line:object-literal-sort-keys
 				builder: {
 					"framework": {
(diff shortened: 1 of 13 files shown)


lib/commands/generate.ts
```diff
@@ -7,15 +7,15 @@ import { default as config } from "./config";
 const command = {
 	aliases: ["g"],
 	command: "generate",
-	desc: "Generates new project",
+	desc: "generates custom template",
```
damyanpetev: I guess this can stay for now, though `generate` can later on be extended beyond just custom templates

lib/commands/new.ts
```diff
@@ -99,6 +99,12 @@ command = {
 		if (!argv["skip-git"] && !ProjectConfig.getConfig().skipGit) {
 			Util.gitInit(process.cwd(), argv.name);
 		}
+
+		Util.log("");
+		Util.log("Next Steps:");
```
damyanpetev: 🙌

Issue #66, opened 2017-11-15
Add doc command


damyanpetev:
Add a `ignite-ui <search term>` command that would open the Online help search for the defined term.

PR #209, opened 2018-03-02
feat: add doc command #66

ViktorSlavov:
#66 

Add doc command, allowing users to search the infragistics help for a given term

### DIFF
--- a/lib/PromptSession.ts
+++ b/lib/PromptSession.ts
@@ -10,6 +10,20 @@ export class PromptSession {
 
 	constructor(private templateManager: TemplateManager) { }
 
+	public static async chooseTerm() {
+		const answers = await inquirer.prompt({
+			default: null,
+			message: "Enter a search term",
+			name: "term",
+			type: "input"
+		});
+		if (answers.term) {
+			return answers.term;
+		} else {
+			const retProm = await this.chooseTerm();
+			return retProm;
+		}
+	}
 	/**
 	 * Start questions session for project creation
 	 */
--- a/lib/commands/doc.ts
+++ b/lib/commands/doc.ts
@@ -1,0 +1,39 @@
+import * as opn from "opn";
+import { PromptSession } from "../PromptSession";
+import { Util } from "../Util";
+
+let doc: {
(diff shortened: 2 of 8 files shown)


lib/PromptSession.ts
```diff
@@ -265,3 +265,20 @@ export class PromptSession {
 		return answers;
 	}
 }
+
+export function chooseTerm() {
+	return inquirer.prompt({
```
damyanpetev: you can use `await` to cut down on promise handlers :)
`const answers = await inquirer.prompt(...```

lib/PromptSession.ts
```diff
@@ -265,3 +265,20 @@ export class PromptSession {
 		return answers;
 	}
 }
+
+export function chooseTerm() {
```
damyanpetev: And this one should be marked as `async` really (though await will force it anyway)
Just for style sake, I'd prefer if the extra function was part of the class itself (matching exported class to file name thing), could be static.

lib/commands/doc.ts
```diff
@@ -0,0 +1,42 @@
...
+				return this.execute(argv);
+			});
+		} else if (!Util.isAlphanumericExt(argv.term)) {
+			return Util.error(`The search term '${argv.term}' is not valid.` + "\n" +
+"Name should start with a letter and can also contain numbers, dashes and spaces.",
+			"red");
+		} else {
+			Util.error(`Review your search results in the browser`, "green");
+			return (argv.extended) ?
+			doc.open(`https://www.google.com/search?q=site%3Aigniteui.com\\%20infragistics.com+${argv.term}`) :
```
damyanpetev: One too many?

lib/commands/doc.ts
```diff
@@ -0,0 +1,42 @@
...
+			type: "string"
+		}
+	},
+	template: null,
+	open(target) {
+		opn(target);
+	},
+	async execute(argv) {
+		if (!argv.term) {
+			await chooseTerm().then(answer => {
```
damyanpetev: no need for then if you use `await`

PR #219, opened 2018-03-07
feat: add validation for configuration options in config set command

### LINKED ISSUE #162: Configuration options validation and type checks
A continuation of #67, logging for future improvements.
If we are to expand the `config` command(s) beyond the currently listed two options that made sense for now, I believe we'll need some tweaks:
- Check strictly for options against the Config interface or allow custom properties
- Type checks and cast for non-strings (meybe using something like [typescript-json-schema](https://github.com/YousefED/typescript-json-schema))
- Get/set of nested options?

wnvko:
Closes #162

### DIFF
--- a/lib/commands/config.ts
+++ b/lib/commands/config.ts
@@ -79,11 +79,12 @@ const command = {
 			config = ProjectConfig.localConfig();
 		}
 
-		if (!ProjectConfig.validateProperty(argv.property, argv.value)) {
+		const validationResult = ProjectConfig.validateProperty(argv.property, argv.value);
+		if (!validationResult.valid) {
 			return;
 		}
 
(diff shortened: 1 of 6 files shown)


lib/commands/config.ts
```diff
@@ -116,6 +116,44 @@ const command = {
 		config[argv.property].push(argv.value);
 		ProjectConfig.setConfig(config, argv.global);
 		Util.log(`Property "${argv.property}" updated.`);
+	},
+	validateProperty(property, value, schemaPath): boolean {
+		//	TODO: check if schema path is actually existing file
+		const schema = require(schemaPath);
```
damyanpetev: You can probably move this to the config module (easier to mock for tests)

PR #243, opened 2018-04-20
Add npm package dependencies support for templates

### LINKED ISSUE #192: Support npm dependencies into CLI template
As a developer I need to be able to provide npm package as a dependency for my template/custom template

damyanpetev:
Closes #192

### DIFF
--- a/lib/types/Template.d.ts
+++ b/lib/types/Template.d.ts
@@ -8,6 +8,8 @@ declare interface Template extends BaseTemplate {
 	/** Step by step */
 	listInComponentTemplates: boolean;
 	listInCustomTemplates: boolean;
+	/** Extra packages to install when adding to project */
+	packages: string[];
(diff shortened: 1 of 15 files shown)


lib/types/Template.d.ts
```diff
@@ -8,6 +8,8 @@ declare interface Template extends BaseTemplate {
 	/** Step by step */
 	listInComponentTemplates: boolean;
 	listInCustomTemplates: boolean;
+	/** Extra packages to install when adding to project */
+	packages: string[];
```
bazal4o: We should update Extensibility Wiki (the table).
Also we need to update generate template in all project libraries with the packages.

damyanpetev: Updated https://github.com/IgniteUI/igniteui-cli/wiki/Extensibility-Support#configure-templatejson

PR #283, opened 2018-05-21
fix: Add custom parameters to Google Analytics

### LINKED ISSUE #282: Generate analytics information for the chosen framework
### Description  
We should have separate analytics for the framework end user has chosen while using IgniteUI CLI.

wnvko:
Closes #282


lib/commands/config.ts
```diff
@@ -76,6 +76,14 @@ const command = {
 			return;
 		}
 		const config = ProjectConfig.getConfig(argv.global);
+
+		GoogleAnalytics.post({
+			cd: "Config",
```
damyanpetev: "Config" as screen view alone isn't too meaningful as you can't tell apart the get from the set.

lib/commands/test.ts
```diff
@@ -21,15 +21,28 @@ const command = {
...
 		}
 
-		command.test(argv);
-	},
-	async test(argv) {
 		const projConfig = ProjectConfig.getConfig().project;
+
+		if (!argv.skipAnalytics) {
+			GoogleAnalytics.post({
+				cd: "Build",
```
damyanpetev: Test?

lib/commands/config.ts
damyanpetev: I guess the same applies to generate + template, though I'm ot sure what param can be used to specify sub command and args easily. Maybe further details can be attached via events as far as I can tell or just separate screen names if nothing else works.

PR #294, opened 2018-07-12
Add back option where necessary in the wizard, closes #293

### LINKED ISSUE #293: Wizard doesn't let you go back
### Description  
The wizard doesn't let you go back when you make a choice. Please add entry for back.

- Ignite UI CLI version: 1.3.3
- Framework: Angular
- Project type: Ignite UI for Angular

wnvko:
Closes #293 

We have now `back` option in the wizard. As soon as you reach `Choose Action Step` you will be able to return from `Add component` and `Add view` actions. Also if one choose `Add component` action he will be able to return `Choose group` sub action.

We do not have `back` option on `Choose framework` step. We do not think we should add it.


lib/PromptSession.ts
```diff
@@ -188,101 +189,120 @@ export class PromptSession {
...
+
+					// runTemplateCollection (item: Template[])
+					//TODO refactor
+					const templates: Template[] = pickedComponent.templates;
+					if (templates.length === 1) {
+						//get the only one template
+						selectedTemplate = templates[0];
+					} else {
+						const templateNames = templates.map(x => x.name);
+						const template = await inquirer.prompt({
```
damyanpetev: There should also be an option to go back from here ( Components > Grid > List of grid templates)

PR #297, opened 2018-07-12
Change the date column values with real dates, #290

### LINKED ISSUE #290: Update sample data in Angular Grid template
Update the sample data of the Grid template. 'date' columns should have Date values similar to the Custom Grid template.

- Ignite UI CLI version: 6.0.0
- Framework - Angular

wnvko:
Closes #290

The date column values are now real dates instead of strings. Also, we are handling onColumnInit event and we are setting a formatter on the date column


templates/angular/igx-ts/grid/default/files/src/app/__path__/__name__.component.ts
```diff
@@ -14,4 +15,10 @@ export class $(ClassName)Component implements OnInit {
   ngOnInit() {
     this.localData = employeesData;
   }
+
+  public onColumnInit(column: IgxColumnComponent) {
+    if (column.field === 'RegistererDate') {
+      column.formatter = (date => date.toDateString());
```
damyanpetev: I'd use `toLocaleDateString` here - it's localized and usually shorter format that fits better in this particular grid :)

wnvko: Done!

PR #300, opened 2018-07-13
Command - Add - Can now choose in which module should template be added

### LINKED ISSUE #273: Support add component to a specific module
As I developer I want to be able to choose (step-by-step ?)/provide (command) - module where I want to add the view. 
It could be `add <template> <name> --module=myCustomModule`
similar to `ng g component nameComponent --module=app.module.ts`

ViktorSlavov:
#273


damyanpetev:
Yup, this looks like about what I imagined the change to be :) Keep it up 👍

ViktorSlavov:
We need to add functionality that automatically appends the template to the specified module's export declaration. 
If a shared module declares a component that is later to be used in app-root, the shared module needs to also export it.

lib/project-utility/TypeScriptFileUpdate.ts
```diff
@@ -121,20 +122,22 @@ export class TypeScriptFileUpdate {
...
+	public addDeclaration(filePath: string, addToExport?: boolean) {
 		let className: string;
 		const fileSource = TsUtils.getFileSource(filePath);
 		const relativePath: string = TsUtils.relativePath(this.targetPath, filePath, true, true);
-
 		className = TsUtils.getClassName(fileSource.getChildren());
-
-		this.addNgModuleMeta({ declare: className, from: relativePath });
+		if (addToExport) {
+			this.addNgModuleMeta({ declare: className, from: relativePath }, null, addToExport);
```
damyanpetev: You already have `export` property in `TemplateDependency`, so why not pass in:
```js
{ declare: className, export: className, from: relativePath }
```

lib/templates/IgniteUIForAngularTemplate.ts
```diff
@@ -8,12 +8,13 @@ import { AngularTemplate } from "./AngularTemplate";
 export class IgniteUIForAngularTemplate extends AngularTemplate {
 
 	public dependencies: TemplateDependency[] = [];
+	public framework: string = "angular-wrappers";
```
damyanpetev: nope

lib/commands/add.ts
```diff
@@ -88,12 +95,17 @@ command = {
...
 
-			await command.addTemplate(argv.name, selectedTemplate);
+			await command.addTemplate(argv.name, selectedTemplate, argv.module);
 			await PackageManager.flushQueue(true);
 			PackageManager.ensureIgniteUISource(config.packagesInstalled, command.templateManager);
 		}
 	},
-	async addTemplate(name: string, template: Template): Promise<boolean> {
+	async addTemplate(name: string, template: Template, modulePath?: string): Promise<boolean> {
+		if (modulePath && !Util.directoryExists(path.join(process.cwd(), modulePath))) {
```
damyanpetev: @wnvko This shouldn't really work - modulePath is supposed to point to a file so a directory check should always fail, no?
Bonus - the path is relative to the structure inside the project (e.g. "src/app/modulePath") so you can't quite check it against the current dir. I guess we can add it as an option to `generateFiles` (last optional parameter) so templates can verify it with the correct path.

PR #306, opened 2018-07-18
Added routing for incorrectly typed address and for uncaught error.

### LINKED ISSUE #270: Add global error page handler to the App host project
As a developer I expect when I create a project with igniteui-cli to have set up global error page handler. The project should include simple error page where the handler will navigate to in case of error. 
https://angular.io/api/core/ErrorHandler

jackofdiamond5:
Closes #270

### DIFF
--- a/lib/project-utility/TypeScriptFileUpdate.ts
+++ b/lib/project-utility/TypeScriptFileUpdate.ts
@@ -95,9 +95,16 @@ export class TypeScriptFileUpdate {
 					const newObject = ts.createObjectLiteral([routePath, routeComponent, routeData]);
 					this.createdStringLiterals.push(linkPath, linkText);
 
+					const notFoundWildCard = "**";
+					const nodes = ts.visitNodes(array.elements, visitor);
+					const errorRouteNode = nodes.filter(element => element.getText().includes(notFoundWildCard))[0];
+					const resultNodes = nodes
+						.slice(0, nodes.indexOf(errorRouteNode))
+						.concat(newObject)
+						.concat(errorRouteNode);
+
 					const elements = ts.createNodeArray([
-						...ts.visitNodes(array.elements, visitor),
-						newObject
+						...resultNodes
 					]);
 
 					return ts.updateArrayLiteral(array, elements);
--- a/lib/project-utility/TypeScriptFileUpdate.ts
+++ b/lib/project-utility/TypeScriptFileUpdate.ts
@@ -238,10 +238,6 @@ export class TypeScriptFileUpdate {
 			return;
 		}
 
-		const targetSource = this.targetSource.statements.slice(0, this.importsMeta.lastIndex);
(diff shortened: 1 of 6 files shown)


templates/angular/igx-ts/projects/empty/files/src/app/error-routing/error-handle/global-error-handler.ts
```diff
@@ -0,0 +1,26 @@
+import { ErrorHandler, EventEmitter, Injector, NgZone, Injectable, isDevMode } from '@angular/core';
```
wnvko: I would call this file global-error-handler.service.ts

templates/angular/igx-ts/projects/empty/files/src/app/app.module.ts
```diff
@@ -1,17 +1,22 @@
...
 import { HomeComponent } from './home/home.component';
+import { PageNotFoundComponent } from './error-routing/page-not-found/page-not-found.component'; 
+import { GlobalErrorHandlerService } from './error-routing/error-handle/global-error-handler';
+import { UncaughtErrorComponent } from './error-routing/error-handle/uncaught-error.component';
 
 @NgModule({
   declarations: [
     AppComponent,
-    HomeComponent
+	HomeComponent,
```
damyanpetev: Indent formatting - there are some tabs where spaces should be 😉

damyanpetev:
Looks like a solid first go 🎉

I have to agree with @wnvko on the naming convention, and I'd suggest some more adjustments to the folders:
```
app/error-routing/
├── error/
│   └── ...
└── not-found/
    └── not-found.component.ts
```
Just to keep the paths under control. In fact, I think you can add an `error-routing.module.ts` in that folder to declare the error view components and import+export it through the routing module to keep them out of the main app module :)

lib/project-utility/TypeScriptFileUpdate.ts
```diff
@@ -95,9 +95,16 @@ export class TypeScriptFileUpdate {
 					const newObject = ts.createObjectLiteral([routePath, routeComponent, routeData]);
 					this.createdStringLiterals.push(linkPath, linkText);
 
+					const notFoundWildCard = "**";
+					const nodes = ts.visitNodes(array.elements, visitor);
+					const errorRouteNode = nodes.filter(element => element.getText().includes(notFoundWildCard))[0];
+					const resultNodes = nodes
+						.slice(0, nodes.indexOf(errorRouteNode))
+						.concat(newObject)
+						.concat(errorRouteNode);
```
damyanpetev: @jackofdiamond5 The failing tests turned out to be right - we missed that one. If you encounter a routes array without the "**" wildcard, the `errorRouteNode` will be `undefined` and adding it to the array results in `[node, node, undefined]` which `ts.updateArrayLiteral` won't handle as it expects all valid nodes.

PR #315, opened 2018-07-20
fix(npm): update browsersync dep, drop lite, update package-lock

damyanpetev:
related #300


damyanpetev:
Obsoleted by https://github.com/IgniteUI/igniteui-cli/pull/394

PR #326, opened 2018-07-24
Templates - Angular - Add links to feature docs on custom IgxGrid template

ViktorSlavov:
Enabled features in the IgxGrid custom template will now be listed as links and lead to the resp. feature's documentation page.


wnvko:
When no features are selected above the grid this text appears:

Active features:

It should not be there.

PR #332, opened 2018-08-02
Command - Add - Add to custom module accepts module through alias ('m')

ViktorSlavov:
Custom module for add command can be passed under -m, e.g. `-m=module-path/module-name.module.ts`

### DIFF
--- a/lib/commands/add.ts
+++ b/lib/commands/add.ts
@@ -94,8 +94,8 @@ command = {
 				cd11: !!config.skipGit,
 				cd14: config.project.theme
 			});
-
-			await command.addTemplate(argv.name, selectedTemplate, argv.module);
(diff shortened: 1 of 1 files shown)


lib/commands/add.ts
```diff
@@ -94,8 +94,8 @@ command = {
 				cd11: !!config.skipGit,
 				cd14: config.project.theme
 			});
-
-			await command.addTemplate(argv.name, selectedTemplate, argv.module);
+			const selectedModule = argv.module || argv.m;
```
damyanpetev: This shouldn't be needed - yargs will assign both the alias (`args.m`) and the param (`args.module`) if one is provided.

ViktorSlavov:
Turns out this already works on master, closing PR

PR #348, opened 2018-08-09
feat(calendar): Adding a template for igxCalendar #333

### LINKED ISSUE #333: Add calendar template

dafo:
Closes #333  .  

Adding a template for igxCalendar. 
Should be tested for all the select types: single, multi and range.


templates/angular/igx-ts/calendar/default/files/src/app/__path__/__name__.component.html
```diff
@@ -0,0 +1,8 @@
+<div igxLayout igxLayoutDir="columns" igxLayoutItemAlign="center">
+	<p>igx-calendar component</p>
```
bazal4o: Let's have a placeholder here which clarifies what is the selection mode currently displayed. You car pass the value of the placeholder when setting the extra configuration.

PR #379, opened 2018-10-22
Adding the Batch Updating template 

dafo:
This is still WIP!

Do not merge!


PlamenaMiteva:
The grid does not spread over the whole container when run for the very first time. However, it does if the browser is refreshed.

dafo:
> The grid does not spread over the whole container when run for the very first time. However, it does if the browser is refreshed.

@PlamenaMiteva This issue is addressed with my last push. You can test it again. Thanks!

dafo:
The codacy errors should be fixed

PR #380, opened 2018-10-25
Do not push to GA on non-global install and uninstal

wnvko:
Up to now we were sending GA information on each install and uninstall. However, as each call to
`npm install`
in project containing igniteui-cli forces starting of install and uninstall scripts in non-global mode, we decided to start these scripts only when installation is global.

### DIFF
--- a/scripts/install.ts
+++ b/scripts/install.ts
@@ -1,11 +1,13 @@
 import { GoogleAnalytics } from "../lib/GoogleAnalytics";
 
 function install() {
-	GoogleAnalytics.post({
-		t: "screenview",
-		// tslint:disable-next-line:object-literal-sort-keys
-		cd: `install global: ${!!process.env.npm_config_global}`
-	});
+	if (!!process.env.npm_config_global) {
+		GoogleAnalytics.post({
+			t: "screenview",
+			// tslint:disable-next-line:object-literal-sort-keys
+			cd: `install`
+		});
+	}
 }
 
 install();
--- a/scripts/uninstall.ts
+++ b/scripts/uninstall.ts
@@ -1,11 +1,13 @@
 import { GoogleAnalytics } from "../lib/GoogleAnalytics";
 
 function uninstall() {
-	GoogleAnalytics.post({
-		t: "screenview",
-		// tslint:disable-next-line:object-literal-sort-keys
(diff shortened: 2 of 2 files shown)


scripts/install.ts
```diff
@@ -1,11 +1,13 @@
...
 function install() {
-	GoogleAnalytics.post({
-		t: "screenview",
-		// tslint:disable-next-line:object-literal-sort-keys
-		cd: `install global: ${!!process.env.npm_config_global}`
-	});
+	if (!!process.env.npm_config_global) {
+		GoogleAnalytics.post({
+			t: "screenview",
+			// tslint:disable-next-line:object-literal-sort-keys
```
kdinev: Why don't you just sort the two keys, instead of disabling the line?

scripts/uninstall.ts
```diff
@@ -1,11 +1,13 @@
...
 function uninstall() {
-	GoogleAnalytics.post({
-		t: "screenview",
-		// tslint:disable-next-line:object-literal-sort-keys
-		cd: `uninstall global: ${!!process.env.npm_config_global}`
-	});
+	if (!!process.env.npm_config_global) {
+		GoogleAnalytics.post({
+			t: "screenview",
+			// tslint:disable-next-line:object-literal-sort-keys
```
kdinev: Why don't you just sort the two keys, instead of disabling the line?

wnvko: We are listing post() parameters everywhere in same order:

1. `t` - type of post;
2. `cd` - screen name, or `ec` - event category;
3. `el` - event label, if applicable;
4. `ea` -event action, if applicable;
5. `cd[1..12]` - custom dimensions if any.

This is why we have these two in this order.

wnvko: Finally we decided to change tslint config and added:

`"object-literal-sort-keys": [ true, "ignore-case", "match-declaration-order", "shorthand-first" ]`

This will force lint to check object's keys order against its interface if any. This allows us to remove 

`// tslint:disable:object-literal-sort-keys`

line from most of the code.

bazal4o:
Moving directly to status verified, because building the package and installing it on machine of the internal network won't generated any GA record

PR #390, opened 2018-11-06
Update ignite-ui-cli.json to use correct project template

bazal4o:


damyanpetev:
@bazal4o Actually, since those projects lack migration mechanism, out handling should probably be compatible with older configs as well

PR #398, opened 2018-11-08
react: added sample url for running tests

jackofdiamond5:


damyanpetev:
@jackofdiamond5 Can you re-target the PR to 2.1.x so the fix can be released for that version too?

PR #404, opened 2018-11-13
feat(igxGrid): Updating the batch editing template

dafo:
Adding a commit dialog to the template


PlamenaMiteva:
The tests fail because the GridWithTransactionsComponent should also be imported and the double quotes in the grid-transaction-component should be converted to single ones.

damyanpetev:
@dafo @PlamenaMiteva There's a weird behavior when you scroll the transactions grid and close it like that (cancel). When you add another transaction (delete/add/..) and open the dialog again the scrollbar has returned to it's original position, but the grid table view hasn't. Not sure that's related to the sample, but worth investigating.

PR #425, opened 2018-12-13
Fix CustomDateSummary result

### LINKED ISSUE #424: grid-summaries template does not render because of empty data
### Description  
grid-summaries template does not render because toLocaleDateString() call on empty data.

- Ignite UI CLI version: 3.2.x
- Framework: Angular
- Project type: igx-ts

### Result  
The "grid-summaries" template generated via the cli does not render.

### Expected result  
The "grid-summaries" template should render.

IvayloG:
Closes #424


templates/angular/igx-ts/grid/grid-summaries/files/src/app/__path__/__name__.component.ts
```diff
@@ -13,7 +13,7 @@ import { employeesData } from './localData';
 export class GridSummariesComponent implements OnInit {
```
bazal4o: The classname should be templated. -> $(ClassName)Component

if you add two template with any name the build will fail.

PR #434, opened 2019-01-10
Add a more descriptive message to an igx-ts/projects/polyfills.ts import

jackofdiamond5:


templates/angular/igx-ts/projects/_base/files/src/polyfills.ts
```diff
@@ -47,7 +47,10 @@ import 'core-js/es6/set';
 /** IE10 and IE11 requires the following for the Reflect API. */
  import 'core-js/es6/reflect';
 
-/** Adding ES7 polyfills for Object for IE. */
+/**
+ *  Adding ES7 polyfills to allow igxGrid to render in IE.
```
damyanpetev:
```suggestion
 *  ES7 `Object.entries` needed for igxGrid to render in IE. 
```

PR #472, opened 2019-03-05
Update input group template styling 

### LINKED ISSUE #471: Input group time picker wrapping container styling doesn't match the form
### Description  
Input group time picker wrapping container styling doesn't match the form. It's using `igx-time-picker` class which is actually used in the product theming. The template shouldn't be using this class for no reason and this also applies to the combo and date picker.

- Ignite UI CLI version: 4.0.0 (reproducible in lower versions too)
- Framework: Angular
- Project type: igx-ts

### Steps to reproduce  

1. Create Ignite UI for Angular project
2. Add input group (Add component > Data Entry & Display > Input Group)
3. Run the app

### Result  
Input group time picker looks odd - shorter than the other fields in the form and with shadow:
image

### Expeced result  
Time picker should match the rest of the fields.

### Attachments  
Attach a sample if available, and screenshots, if applicable.

Lipata:
Closes #471


Lipata:
The release of the igniteui-angular is needed to test if select works without the DropDownModule.

PR #489, opened 2019-03-27
fix: Side-nav-auth project to work correctly with payload encoded in UTF-8

wnvko:
We were using old version of `angular-auth-oidc-client` (^6.0.6) where decoding of payload did not support non ASCII symbols, e.g. Cyrillic or Greek letters. This was fixed in most recent versions and we are switching now to currently latest version ^9.0.3.

Additionally we add support to payload with non ASCII symbols, e.g. Cyrillic or Greek letters, in our module responsible for authentication.


wnvko:
`JWT` standard requires sequence of URL-safe parts separated by period ('.') characters.  Each part contains a `base64url`-encoded value. We are using `window.bota` function to encode the each part of `JWT`. However, `btoa` works only with UTF-8 characters from Latin-1 range. When user provide personal info in UTF-8 outside Latin-1 range, e.g. Arabic or Japanese name, `btoa` fails. To prevent this we are escaping both `header` and `playload` parts of `JWT` by calling `encodeURI` over each part.
Calling `encodeURI` does not return JWS Compact Serialization as [required by the standard](https://tools.ietf.org/html/rfc7519#section-3). This is why as the last step, before calling `btoa`, we are unescaping the result of `encodeURI` converting each escaped symbol back to one in UTF-8 Latin-1 range. Last step is similar to what deprecated `window.unescape` method will generate from same input.

templates/angular/igx-ts/projects/side-nav-auth/files/src/app/authentication/services/fake-backend.service.ts
```diff
@@ -124,11 +125,11 @@ export class BackendInterceptor implements HttpInterceptor {
...
     /** Note: This is used for example purposes only and does NOT generate a valid JWT w/ base64Url encoding */
     private generateToken(payload: any): string {
-        const header = JSON.stringify({
+        const header = {
             alg: 'Mock',
             typ: 'JWT'
-        });
-        return btoa(JSON.stringify(header)) + '.' + btoa(JSON.stringify(payload)) + '.mockSignature';
+        };
+        return btoa(encodeBase64Url(header)) + '.' + btoa(encodeBase64Url(payload)) + '.mockSignature';
```
damyanpetev: Either `encodeBase64Url` does the actual `btoa` base 64 encoding.. or it should probably be named something like encode non-ASCII chars or something

PR #490, opened 2019-03-27
fix(igx-ts): login view scenario is properly templated

ViktorSlavov:


templates/angular/igx-ts/custom-templates/login/files/src/app/__path__/__name__.component.scss
```diff
@@ -1,4 +1,5 @@
 :host {
+	display: block;
```
damyanpetev:
```suggestion
    display: block;
```

damyanpetev:
Just need `NoopAnimationsModule` in the test file

PR #496, opened 2019-04-15
Replace ambient types with exports

jackofdiamond5:


lib/types/GoogleAnalytics.ts
```diff
@@ -0,0 +1,4 @@
+export interface GoogleAnalytics {
```
damyanpetev: I'm not sure this is even used anywhere, actually

PR #509, opened 2019-05-10
refactor(*): separate processTemplate / validateTemplate from Template Class

ViktorSlavov:
template.generateFiles now returns the config for the template

`processTemplate()` and `validateTemplate()` are called separately (with the newly generated config), instead of in `generateFiles()`


lib/types/BaseTemplate.ts
```diff
@@ -22,9 +22,11 @@ export interface BaseTemplate {
 
 	/** This property controls if extra configuration is available to the template */
 	hasExtraConfiguration: boolean;
+	/** An array with the physical path to the template files */
+	templatePath: string[];
```
damyanpetev: templatePath**s**

spec/unit/PromptSession-spec.ts
```diff
@@ -180,7 +185,7 @@ describe("Unit - PromptSession", () => {
 
 		expect(Util.log).toHaveBeenCalledWith("  Framework: Custom Framework 1");
 		expect(Util.log).toHaveBeenCalledWith("  Project type: Ignite UI for Angular");
-		expect(mockProject.generateFiles).toHaveBeenCalledWith(process.cwd(), "Test Project", mockProjectLibrary.themes[0]);
+		expect(mockProject.generateConfig).toHaveBeenCalledWith("Test Project", mockProjectLibrary.themes[0]);
```
damyanpetev: The old check for `generateFiles` should now check `processTemplates` has been called correctly

spec/unit/add-spec.ts
```diff
@@ -73,7 +74,7 @@ describe("Unit - Add command", () => {
 
 		for (const item of validCombos) {
 			await addCmd.addTemplate(item.name, mockTemplate);
-			expect(mockTemplate.generateFiles).toHaveBeenCalledWith(process.cwd(), item.valid, jasmine.any(Object));
+			expect(mockTemplate.generateConfig).toHaveBeenCalledWith(item.valid, {});
```
damyanpetev: Same here - also check the `processTemplates` call

PR #514, opened 2019-05-13
Update samples styles

### LINKED ISSUE #513: Select samples do not display ok when in ng project
### Description  
Select samples do not display OK when in ng project. As it is not a project created with igniteui-cli, there are no wrapper styles applied, so it needs additional per-sample styling to display correctly. 
- Ignite UI CLI version: all
- Framework: Angular
- Project type: igx-ts

### Steps to reproduce  

1. Create an empty project using Angular CLI via (ng new newNgProjName)
2. Add select components to the project like "Select with groups", and so on. (ig + wizard) 

### Result  
Group headers are not visible.

### Expected result  
Group headers are should be visible.

### Attachments  
image

IvayloG:
Closes  #513 .

### DIFF
--- a/templates/angular/igx-ts/select/default/files/src/app/__path__/__name__.component.scss
+++ b/templates/angular/igx-ts/select/default/files/src/app/__path__/__name__.component.scss
@@ -1,3 +1,3 @@
 .select-wrapper {
-	padding-top: 50px;
+	max-width: 200px;
 }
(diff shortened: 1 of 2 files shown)


ViktorSlavov:
The width of the `Select with Groups` sample was a bit low and the label was getting skewed:
image
I've expanded it a bit (`200px => 240px`).
Other that than, adjustments work fine.

PR #515, opened 2019-05-17
refactor(templates): Allow template parsing delimiters to be changed

ViktorSlavov:
Closes # .  

Template delimiters are now specified in template declaration and can be changed


lib/templates/AngularTemplate.ts
```diff
@@ -17,6 +17,17 @@ export class AngularTemplate implements Template {
 	public projectType: string;
 	public hasExtraConfiguration: boolean = false;
 	public packages = [];
+	public delimiters = {
```
damyanpetev: = Util.defaultDelimiters

lib/templates/IgniteUIForAngularTemplate.ts
```diff
@@ -64,9 +75,11 @@ export class IgniteUIForAngularTemplate extends AngularTemplate {
 				// relative file dependency
 				const copy = Object.assign({}, dep);
 				copy.from = Util.relativePath(mainModulePath, path.join(projectPath, copy.from), true, true);
-				mainModule.addNgModuleMeta(copy, this.getBaseVariables(name));
+				mainModule.addNgModuleMeta(copy,
+					Util.applyDelimiters(this.getBaseVariables(name), this.delimiters.content || defaultDelimiters.content));
```
damyanpetev: The defaults check is not needed, considering the property is defined above.

PR #552, opened 2019-08-28
refactor(*): move CLI, add lerna, automate deployment

ViktorSlavov:
Closes # .  


packages/core/util/Util.ts
```diff
@@ -304,12 +304,15 @@ export class Util {
 				return process.exit();
 			}
 
-			// if SIGINT killed the process with no errors
-			if (error.status === 3221225786 || error.status > 128) {
-				return process.exit();
+			// we might want to skip dropping the whole process chain on error codes
+			// in the case of chaining w/ npm install
+			if (exitOnError) {
```
ViktorSlavov: @damyanpetev the react quickstart throws errors during `npm install` (which are **not** critical and do not end the process if run in a console manually) - IMO, the easiest fix for the time being (before adjusting/removing the react quickstart portion) is to have a flag on the `execSync` function that forces it to ignore errors. The only place this is called is during the react quickstart. I know it's kinda hacky... Thoughts?

PR #553, opened 2019-09-04
Fix lint errors for generated angular projects

PlamenaMiteva:
Closes # 541.


ViktorSlavov:
There seem to be some files that were missed in the **Authentication** project. Most of them are `.spec` files, which might even be addressed in the test fix PR, but some errors are in the component files - please take a look when you have the time.

There' s also a failing file in the combo template - `local-data.ts`;

PR #586, opened 2019-10-30
fix(schematics): component schematics checks for available names

### LINKED ISSUE #576: [NG schematics] - ng-add and component schematics do not account for virtual file system
### Description  
Currently, the check whether a file (folder) exists is done in `Util` and does not account for the virtual tree when used in the context of the schematics. This leads to calling `ng-add` and `component` not properly prompting a default name and not properly erroring when the name already exist.

- Ignite UI CLI version: 5.0.0
- Framework: Angular|React|jQuery
- Project type: ig-ts|igx-ts

### Steps to reproduce  

1. Run `ng new --collection=@igniteui/angular-schematics`
2. Project Type => Default Side Navigation
3. Add Component => Grid => Grid (do not change the name)
4. Repeat Step 3 several times
5. Complete & Run

### Result  
The prompt does not change the default name to `Grid [increment]`.
In the end, only on Grid component (grid) is created (though routes are added for each instance)

### Expected result  
The prompt changes the default name to `Grid [increment]`.
Separate grid components with the proper names are created in the end.

ViktorSlavov:
Closes #576


packages/ng-schematics/src/component/index.ts
```diff
@@ -72,7 +72,7 @@ export function component(options: ComponentOptions): Rule {
 		const projLib = templateManager.getProjectLibrary("angular", "igx-ts");
 		let prompt: SchematicsPromptSession;
 		if (!options.template || !options.name) {
-			prompt = new SchematicsPromptSession(templateManager, addedComponents);
+			prompt = new SchematicsPromptSession(templateManager, addedComponents, _tree.getDir("src/app").subdirs);
```
ViktorSlavov: Not sure if this is the best way to get the existing folders

PR #601, opened 2019-11-27
ng-new able to run without prompts

### LINKED ISSUE #595: ng-new schematic should be able to run without prompts, respect interactive
### Description  
If enough/all parameters are provided in the command or interactive~/defaults~ flags are used the command should be able to run without going through prompts.

- Framework: Angular

### Steps to reproduce  

For example run 
```bash
 ng new newProject --collection="@igniteui/angular-schematics" --template=side-nav --theme=custom
```
or
```bash
ng new newProject --collection="@igniteui/angular-schematics" --no-interactive
```

### Result  
Regardless what you provide as as options, the new schematics always run through the step by step prompt.

### Expected result  
Prompts should not be displayed if the option is provided or are explicitly disabled.

dafo:
For #595

### DIFF
--- a/packages/ng-schematics/src/ng-new/index.ts
+++ b/packages/ng-schematics/src/ng-new/index.ts
@@ -32,10 +32,6 @@ export function newProject(options: OptionsSchema): Rule {
 		const defaultProjName = "IG Project";
 		let allOptionsProvided: boolean = false;
 
-		if (options.name && options.template && options.theme) {
-			allOptionsProvided = true;
-		}
-
 		return chain([
 			(tree: Tree, _context: IgxSchematicContext): Observable<Tree> => {
 				return defer(async () => {
@@ -47,6 +43,10 @@ export function newProject(options: OptionsSchema): Rule {
 							default: Util.getAvailableName(defaultProjName, true),
 							validate: prompt.nameIsValid
 						});
+					}
+
+					if (options.name && options.template && options.theme) {
+						allOptionsProvided = true;
 					}
 
 					const framework = templateManager.getFrameworkByName("angular");
@@ -112,7 +112,7 @@ export function newProject(options: OptionsSchema): Rule {
 						}
 					},
 					(tree: Tree, context: IgxSchematicContext) => {
-						if (!allOptionsProvided || !options.defaults) {
+						if (!allOptionsProvided) {
 							return defer(async () => {
 								prompt.setContext(context, tree, options.name as string);
 								await prompt.chooseActionLoop(projLibrary);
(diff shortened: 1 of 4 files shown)


dafo:
Need to add support for commands like
`ng new newProject --collection="@igniteui/angular-schematics" --defaults --theme=custom`
Also, need to fix failing tests. However, a review of the functionality itself may be done at this point.

dafo:
Still need to fix the tests.

packages/ng-schematics/src/ng-new/index.ts
```diff
@@ -100,11 +112,13 @@ export function newProject(options: OptionsSchema): Rule {
 						}
 					},
 					(tree: Tree, context: IgxSchematicContext) => {
-						return defer(async () => {
-							prompt.setContext(context, tree, options.name as string);
-							await prompt.chooseActionLoop(projLibrary);
-							return tree;
-						});
+						if (!allOptionsProvided || !options.defaults) {
```
ViktorSlavov: I think this check should just be `!allOptionsProvided`
If you've provided `theme`, `name` & `template`, it's still possible that `!option.defaults === true` and this still enters.
If `options.defaults === true`, then `allOptionsProvided === true`, so it won't enter

packages/ng-schematics/src/ng-new/index.ts
```diff
@@ -30,6 +30,11 @@ export function newProject(options: OptionsSchema): Rule {
 
 		// TODO:
 		const defaultProjName = "IG Project";
+		let allOptionsProvided: boolean = false;
+
+		if (options.name && options.template && options.theme) {
```
ViktorSlavov: @damyanpetev @dafo 
Maybe we should move this under the `name` prompt?
Currently, if you call the schematic like so:
```
ng new --collection=igniteui/angular-schematics --theme=Custom --template=side-nav
```
It will prompt you for a name and then go into the action loop.

packages/ng-schematics/src/ng-new/index_spec.ts
```diff
@@ -17,17 +17,6 @@ describe("igniteui-angular-schematics", () => {
 		const runner = new SchematicTestRunner("schematics", collectionPath);
 		const myTree = Tree.empty();
 		const workingDirectory = "my-test-project";
-		const mockLibrary = { getProject: jasmine.createSpy("getProject").and.returnValue(true), themes: ["custom"] };
```
ViktorSlavov: Since a name is provided and `defaults` is `true` by default, there is no longer a need for a mock of the `promptSession`, I agree.
We should add more tests to better cover the scenario though - calling w/o a name, calling w/ `defaults === false` and a combination of the different options to make sure it properly prompts for them (could be done in a separate PR IMO).

packages/ng-schematics/src/ng-new/index.ts
damyanpetev: That's actually better handled than what you'd get with the stand-alone cli equivalent. But technically if you don't provide the name - you ignored the command usage requirements that start the first thing you pass is an argument with name.
I'd accept the partial prompts for just the missing options as an enhancement though, just not for this issue.

PR #602, opened 2019-11-29
Adding a name for schematics projects

### LINKED ISSUE #587: Provide a parameter to differentiate between GA posts from cli package and ng schematics
### Description  
Provide a parameter to differentiate between GA posts from cli package and ng schematics

- Framework: Angular|React|jQuery
- Project type: ig-ts|igx-ts

dafo:
Closes #587

### DIFF
--- a/packages/ng-schematics/src/component/index.ts
+++ b/packages/ng-schematics/src/component/index.ts
@@ -43,7 +43,8 @@ function updatePackageJSON(installContext: { packages: Map<string, string>, shou
 }
 
 function addComponent(options: TemplateOptions, skipRoute = false, modulePath?: string): Rule {
-	return async (_tree, _context) => {
+	return async (tree: Tree, _context: SchematicContext) => {
+		setVirtual(tree);
 		const config = options.templateInst.generateConfig(options.name, {});
 		return chain([...options.templateInst.templatePaths.map(templatePath =>
 			mergeWith(
@@ -55,9 +56,8 @@ function addComponent(options: TemplateOptions, skipRoute = false, modulePath?: 
 	};
 }
 export function singleComponent(templateOptions: TemplateOptions, skipRoute: boolean) {
-	return async (tree: Tree, _context: SchematicContext) => {
+	return async (_tree: Tree, _context: SchematicContext) => {
 		const packages = new Map<string, string>();
-		setVirtual(tree);
 		return chain([
 			addComponent(templateOptions, skipRoute),
 			getMissingPackages(templateOptions, packages),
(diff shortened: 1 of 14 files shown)


packages/ng-schematics/src/component/index.ts
```diff
@@ -67,7 +67,11 @@ export function singleComponent(templateOptions: TemplateOptions, skipRoute: boo
 }
 export function component(options: ComponentOptions): Rule {
 	return async (tree: Tree, context: SchematicContext) => {
-		App.initialize();
+		App.initialize("angular-cli");
+		GoogleAnalytics.post({
```
ViktorSlavov: There is another post in the `ig add` logic that will be skipped here: 
When the user chooses a template, there's a post that [contains information about the choice](https://github.com/IgniteUI/igniteui-cli/blob/vNext/packages/cli/lib/commands/add.ts#L93). This should also be added to the component schematic.

packages/ng-schematics/src/ng-new/index.ts
```diff
@@ -22,7 +22,11 @@ interface IgxSchematicContext extends SchematicContext {
 
 export function newProject(options: OptionsSchema): Rule {
 	return (_host: Tree, _hostContext: IgxSchematicContext) => {
-		App.initialize();
+		App.initialize("angular-cli");
+		GoogleAnalytics.post({
+			t: "screenview",
+			cd: "New"
+		});
 		let projLibrary: ProjectLibrary;
```
ViktorSlavov: There is another post in the `ig new` logic that will be skipped here: 
When the user chooses a project name, framwork and template, there's a post that [contains information about the choice](https://github.com/IgniteUI/igniteui-cli/blob/vNext/packages/cli/lib/commands/new.ts#L115). This should also be added to the `new` schematic.

packages/ng-schematics/src/component/index.ts
```diff
@@ -67,7 +67,11 @@ export function singleComponent(templateOptions: TemplateOptions, skipRoute: boo
...
+			t: "screenview",
+			cd: "Add"
+		});
```
wnvko: At the end, when all custom options are resolved we should call GA again with something like this:
```
			GoogleAnalytics.post({
				t: "event",
				ec: "$ig add", - we may call it here with $ng add
				ea: `template id: ${argv.template}; file name: ${argv.name}`,
				cd1: selectedTemplate.framework,
				cd2: selectedTemplate.projectType,
				cd5: selectedTemplate.controlGroup,
				cd7: selectedTemplate.id,
				cd8: selectedTemplate.name,
				cd11: !!config.skipGit,
				cd14: config.project.theme
			});
```

PR #603, opened 2019-12-03
feat(schematics): skip start task if skipInstall === true

ViktorSlavov:
Related #595 .  

Do not try to run `start` command if `skipInstall` or `allOptionsProvided`


packages/ng-schematics/src/ng-new/index.ts
```diff
@@ -131,8 +131,10 @@ export function newProject(options: OptionsSchema): Rule {
 					installChain.push(gitTask);
 				}
 
+				if (!options.skipInstall && !allOptionsProvided) {
+					context.addTask(new RunSchematicTask("start", { directory: options.name }), installChain);
+				}
 				// TODO: Conditional?
```
damyanpetev:
```suggestion
```

PR #605, opened 2019-12-03
fix(schematics): set virtual FS in proper stage

### LINKED ISSUE #604: Component schematics do not work w/o loop
### Description  
When the components schematic is called inside of a project w/ all properties:
```bash
ng g @igniteui/angular-schematics:component --template=combo --name=bombo
```
The schematics fail

- Ignite UI CLI version: 5.0.0-beta.1
- Framework: Angular
- Project type: igx-ts

### Steps to reproduce  

1. `ng new my-test-project --defaults`
2. `cd my-test-project && npm i @igniteui/angular-schematics@next`
3. `ng g @igniteui/angular-schematics:c grid myGrid`

### Result  
Schematics fail

### Expected result  
Schematics properly generate a grid

ViktorSlavov:
Closes #604.  

The App's fs was never set to virtual when calling the component schematic w/ all options.

This should *not* affect the `ng-new` and `component` w/ prompt options in no way

### DIFF
--- a/packages/igx-templates/IgniteUIForAngularTemplate.ts
+++ b/packages/igx-templates/IgniteUIForAngularTemplate.ts
@@ -121,7 +121,7 @@ export class IgniteUIForAngularTemplate implements Template {
 	}
 	public setExtraConfiguration(extraConfigKeys: {}) { }
 
-	protected getBaseVariables(name: string) {
+	protected getBaseVariables(name: string): { [key: string]: string } {
 		const config = {};
 		config["name"] = Util.nameFromPath(name);
 		config["ClassName"] = Util.className(Util.nameFromPath(name));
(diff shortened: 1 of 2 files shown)


packages/igx-templates/IgniteUIForAngularTemplate.ts
```diff
@@ -121,7 +121,7 @@ export class IgniteUIForAngularTemplate implements Template {
 	}
 	public setExtraConfiguration(extraConfigKeys: {}) { }
 
-	protected getBaseVariables(name: string) {
+	protected getBaseVariables(name: string): { [key: string]: string } {
```
damyanpetev: This can be on a higher level (like the Template interface) if it's not typed there

ViktorSlavov: It's `protected` and not described in the interface. There are no prior inheritance levels where this can be applied, it's a small QoL life 😅. Then again, it's not called anywhere outside of the template itself, so it can be removed.

PR #609, opened 2019-12-06
fix(new, schematics): No prompts when "name" is provided

### LINKED ISSUE #595: ng-new schematic should be able to run without prompts, respect interactive

dafo:
Closes #595.

### DIFF
--- a/packages/ng-schematics/src/ng-new/index.ts
+++ b/packages/ng-schematics/src/ng-new/index.ts
@@ -50,6 +50,10 @@ export function newProject(options: OptionsSchema): Rule {
 						}
 					}
 
+					if (Util.directoryExists(options.name)) {
+						throw new SchematicsException(`Folder "${options.name}" already exists!`);
+					}
+
 					if (!options.name || !prompt.nameIsValid(options.name)) {
 						options.name = await prompt.getUserInput({
 							type: "input",
@@ -150,7 +154,7 @@ export function newProject(options: OptionsSchema): Rule {
 					installChain.push(gitTask);
 				}
 
-				if (!options.skipInstall && options.name) {
+				if (!options.skipInstall && !nameProvided) {
 					context.addTask(new RunSchematicTask("start", { directory: options.name }), installChain);
 				}
 				return tree;
--- a/packages/ng-schematics/src/ng-new/index.ts
+++ b/packages/ng-schematics/src/ng-new/index.ts
@@ -34,11 +34,13 @@ export function newProject(options: OptionsSchema): Rule {
 
 		// TODO:
 		const defaultProjName = "IG Project";
-		let allOptionsProvided: boolean = false;
 
 		return chain([
 			(tree: Tree, _context: IgxSchematicContext): Observable<Tree> => {
 				return defer(async () => {
+					const framework = templateManager.getFrameworkByName("angular");
+					// app name validation???
+					projLibrary = await prompt.getProjectLibrary(framework);
 					if (!options.name || !prompt.nameIsValid(options.name)) {
 						options.name = await prompt.getUserInput({
 							type: "input",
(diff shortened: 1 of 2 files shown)


dafo:
Need to fix the test.  However, the change can be reviewed.

packages/ng-schematics/src/ng-new/index.ts
```diff
@@ -34,11 +34,15 @@ export function newProject(options: OptionsSchema): Rule {
...
+		let nameProvided: boolean = false;
 
 		return chain([
 			(tree: Tree, _context: IgxSchematicContext): Observable<Tree> => {
 				return defer(async () => {
+					if (options.name) {
+						nameProvided = true;
+					}
+
 					if (!options.name || !prompt.nameIsValid(options.name)) {
```
ViktorSlavov: Not sure how this should behave if the provided name is not a valid one. With the current logic, if we pass an invalid name (e.g. `1111`), `nameProvided` will be set to `true` and the schematic will skip the `chooseActionLoop` part.

In the main cli, if an invalid name is provided, the process exits w/ an error message (defined in [new.ts](https://github.com/IgniteUI/igniteui-cli/blob/vNext/packages/cli/lib/commands/new.ts#L74))

packages/ng-schematics/src/ng-new/index.ts
```diff
@@ -49,31 +53,20 @@ export function newProject(options: OptionsSchema): Rule {
...
-					if (!options.theme) {
-						options.theme = await prompt.getTheme(projLibrary);
-					}
-
 					if (options.theme && projLibrary.themes.indexOf(options.theme) === -1) {
 						throw new SchematicsException(`Theme not supported`);
 					}
+					const theme = options.theme || projLibrary.themes[0];
+					const projectTemplate = options.template || projLibrary.projectIds[0];
+					Util.log(`Project Name: ${options.name}, theme ${theme}`);
```
damyanpetev: This can move after the project template check

packages/ng-schematics/src/ng-new/index.ts
```diff
@@ -148,7 +141,7 @@ export function newProject(options: OptionsSchema): Rule {
 					installChain.push(gitTask);
 				}
 
-				if (!options.skipInstall && !allOptionsProvided) {
+				if (!options.skipInstall) {
```
damyanpetev: This should still check for name provided. If you haven't been through the choose action loop, we shouldn't start the project as well

packages/ng-schematics/src/ng-new/index.ts
dafo: I think it is best to let the customer know the provided name is not valid and prompt him to provide a valid name. After a valid name is provided, we will show the Actions loop prompt.

dafo: We have discussed it with @ViktorSlavov  and agreed it should exit the process as it may be invoked by a script and would never execute the prompt.

packages/ng-schematics/src/ng-new/index.ts
```diff
@@ -47,43 +62,34 @@ export function newProject(options: OptionsSchema): Rule {
 							default: Util.getAvailableName(defaultProjName, true),
 							validate: prompt.nameIsValid
 						});
-					}
-
-					if (options.name && options.template && options.theme) {
-						allOptionsProvided = true;
+						nameProvided = false;
```
damyanpetev: This line is not even needed, it's false by default, right?

PR #623, opened 2020-01-07
refactor(templates): rename 'empty-project' to 'empty' template

ViktorSlavov:
Closes #617.  

Rename **Ignite UI for Angular**  `empty-page` template to `empty` to be consistent w/ other frameworks

PR #626, opened 2020-01-09
Prompt for template and theme if no name provided

### LINKED ISSUE #622: ng-new schematic should prompt for theme and template if no name is provided
### Description  
 If not providing a name to the ng new command you should go through all prompts.

Ignite UI CLI version: 5.x.x
Framework: Angular

### Steps to reproduce  

1. Run `ng new --collection="@igniteui/angular-schematics"`

### Result  
You are prompted for a name only. Then you enter the loop. 

### Expected result  
You are prompted for template and theme before reaching the loop.

dafo:
Closes #622.


packages/ng-schematics/src/ng-new/index_spec.ts
```diff
@@ -26,6 +26,10 @@ describe("igniteui-angular-schematics", () => {
 			.and.returnValue(Promise.resolve(true)),
 			getProjectLibrary: spyOn(SchematicsPromptSession.prototype, "getProjectLibrary")
 			.and.returnValue((Promise.resolve(mockLibrary))),
+			getProjectTemplate: spyOn(SchematicsPromptSession.prototype, "getProjectTemplate")
+			.and.returnValue(Promise.resolve("empty-page")),
```
ViktorSlavov: This might fail, because the name of the empty template was was changed to `'empty'` w/ this PR #623 .

dafo: It won`t fail as the projectIds are mocked:
`projectIds: ["empty-page"]`

PR #634, opened 2020-01-14
Create new project when no name but theme is provided

### LINKED ISSUE #608: Exception thrown when executing "ig new" providing a 'theme' but not a 'name'
### Description  
If you run `ig new` providing a theme but not a name it does not prompt for name, template, etc. but throws an exception.

- Ignite UI CLI version: 4.3.2
- Framework: Angular
- Project type: igx-ts

### Steps to reproduce  

1. Execute `ig new --theme=Custom`

### Result  
TypeError: Cannot read property 'trim' of undefined

### Expected result  
You are prompted to provide a name.

dafo:
Closes #608  .  

Test with `ig new` and with the schematics.

### DIFF
--- a/packages/cli/lib/PromptSession.ts
+++ b/packages/cli/lib/PromptSession.ts
@@ -30,7 +30,7 @@ export class PromptSession extends BasePromptSession {
 	/**
 	 * Start questions session for project creation
 	 */
-	public async start() {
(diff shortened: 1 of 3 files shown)


dafo:
If you provide a theme but no name - you should be prompted about the theme as if you didn't provide it.

PR #646, opened 2020-01-22
chore: use latest rc

Lipata:


packages/igx-templates/igx-ts/projects/side-nav-auth/files/package.json
```diff
@@ -33,9 +33,9 @@
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "0.900.0-rc.7",
```
ViktorSlavov: The `@angular-devkit` devDep should also be non-strict
```suggestion
    "@angular-devkit/build-angular": "~0.900.0-rc.0",
```

packages/igx-templates/igx-ts/projects/_base/files/package.json
```diff
@@ -32,9 +32,9 @@
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "0.900.0-rc.7",
```
ViktorSlavov: The `@angular-devkit` dep should also be non-strict
```suggestion
    "@angular-devkit/build-angular": "~0.900.0-rc.0",
```

PR #648, opened 2020-01-27
fix(react): update webpack versions

Lipata:


jackofdiamond5:
I created a react project with the cli and got the following after I ran it:

image

PR #664, opened 2020-02-04
Fix e2e failing tests

IvayloG:
Update template and tests


Lipata:
When creating a project with navigation the test is failing

PR #668, opened 2020-02-04
Fix chip template

IvayloG:
Template breaks the new generated project build.


Lipata:
Closing, because it will be merged from master

PR #738, opened 2020-05-26
Make update command prompts more meaningful

### LINKED ISSUE #737: Make update command prompts more meaningful
### Description  
When calling the new `update-packages` command, users get no indication at the beginning of the packages installation. If running the command /w credentials already in place in `.npmrc`, it looks like the console froze and users might cancel the process.

### Additional Info
The package install from the command can be `--quiet` as well (no need to log the ton of packages/warnings)

dafo:
Closes #737  .  

### DIFF
--- a/packages/cli/lib/commands/upgrade.ts
+++ b/packages/cli/lib/commands/upgrade.ts
@@ -54,7 +54,13 @@ command = {
 					}
 					const success = await project.upgradeIgniteUIPackages(process.cwd(), "");
 					if (success && !argv.skipInstall) {
-						Util.execSync("npm install");
+						Util.log("Please wait while the upgraded npm packages are being installed!");
+						try {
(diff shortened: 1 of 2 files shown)


packages/cli/lib/commands/upgrade.ts
```diff
@@ -54,7 +54,13 @@ command = {
 					}
 					const success = await project.upgradeIgniteUIPackages(process.cwd(), "");
 					if (success && !argv.skipInstall) {
-						Util.execSync("npm install");
+						Util.log("Please wait while the upgraded npm packages are being installed!");
+						try {
+							Util.execSync("npm install --quiet");
+						} catch (error) {
+							Util.log(`Error installing the upgraded npm packages.`);
```
ViktorSlavov, line 61: This could be "Util.error" instead

PR #742, opened 2020-05-27
feat(upgrade): handle multiple licensed packages

### LINKED ISSUE #741: Add support for dock manager in upgrade-packages command in cli and schematics

ViktorSlavov:
Closes #741.  

Handle upgrading imports and dependencies to multiple packages w/ trial and licensed definitions.

### DIFF
--- a/packages/igx-templates/Update.ts
+++ b/packages/igx-templates/Update.ts
@@ -1,10 +1,36 @@
 import { App, FS_TOKEN, IFileSystem, PackageManager, ProjectConfig, Util } from "@igniteui/cli-core";
 import * as path from "path";
-import { FEED_PACKAGE, NPM_PACKAGE, resolvePackage } from "./package-resolve";
+import { FEED_DOCK_MANAGER, FEED_PACKAGE, NPM_DOCK_MANAGER, NPM_PACKAGE, resolvePackage } from "./package-resolve";
 
 const styleExtension = ["css", "scss", "sass"];
-const styleExpression = String.raw`(node_modules\/|~)${NPM_PACKAGE}(\/)`;
-const logicExpression = String.raw`(from ["'])${NPM_PACKAGE}(["'])`;
+enum RegularExpressionType {
+	STYLE,
+	LOGIC
+}
+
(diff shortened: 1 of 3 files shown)


packages/igx-templates/Update.ts
```diff
@@ -1,10 +1,36 @@
...
+function createExpression(expressionType: RegularExpressionType, packageName: string): string {
+	if (expressionType === RegularExpressionType.LOGIC) {
+		return String.raw`(from ["'])${packageName}(?<submodules>\/.*?)?(["'])`;
+	} else if (expressionType === RegularExpressionType.STYLE) {
+		return String.raw`(node_modules\/|~)${packageName}(\/)`;
+	}
+}
 
 export async function updateWorkspace(rootPath: string): Promise<boolean> {
 	if (resolvePackage() === FEED_PACKAGE) {
```
damyanpetev, line 36: This + the folder check for `@infragistics` are shaky at best, but will have to do for initial testing, especially now with the dock manager added. Will need a bit of re-work to handle both packages (separate resolves) and perhaps respect the root path.

packages/igx-templates/Update.ts
```diff
@@ -1,10 +1,36 @@
...
+		trial: NPM_DOCK_MANAGER,
+		licensed: FEED_DOCK_MANAGER,
+		styleExp: createExpression(RegularExpressionType.STYLE, NPM_DOCK_MANAGER),
+		logicExp: createExpression(RegularExpressionType.LOGIC, NPM_DOCK_MANAGER)
+	}
+};
+
+function createExpression(expressionType: RegularExpressionType, packageName: string): string {
+	if (expressionType === RegularExpressionType.LOGIC) {
+		return String.raw`(from ["'])${packageName}(?<submodules>\/.*?)?(["'])`;
```
damyanpetev, line 29: This is potentially a bit too complicated for the needs - the ending can be limited by a simple look-ahead for a quote or slash - e.g. `(from ["'])igniteui-angular(?=['"\/])` so you won't have an optional capture group

PR #772, opened 2020-06-18
fix: category reorder

Lipata:


Lipata:
image

Let's leave it that way for the moment. It can be changed wherever we decide to do it.

PR #776, opened 2020-07-01
fix: set Angular as default framework choice

Lipata:


IvayloG:
quickstart - generates **ig-ts** type of quickstart. Is this expected because of the existing code base or should it default to **igx-ts** project type as well?

PR #777, opened 2020-07-01
feat(tsutil): add optional parent route to TsFileUpdateUtil.addRoute method

ViktorSlavov:
Add the option for `TsFileUpdateUtil.AddRoute` to add route entries as children of an existing route path.
If `parentRoutePath` parameter (new, optional) is passed, the generated route entry will be added in the `children` array of the route matching the provided path.

### DIFF
--- a/packages/core/typescript/TypeScriptFileUpdate.ts
+++ b/packages/core/typescript/TypeScriptFileUpdate.ts
@@ -5,7 +5,7 @@ import { FS_TOKEN, IFileSystem } from "../types/FileSystem";
 import { Util } from "../util/Util";
 import { TypeScriptUtils as TsUtils } from "./TypeScriptUtils";
 
-const ROUTES_VARIABLE = "routes";
+const DEFAULT_ROUTES_VARIABLE = "routes";
 declare type NodeVisitor = (node: ts.Node) => ts.Node;
 /**
  * Apply various updates to typescript files using AST
@@ -69,11 +69,16 @@ export class TypeScriptFileUpdate {
 	 * @param parentRoutePath __Optional__ If passed, will include the new route as a **child** of the specified route path
 	 */
 	public addRoute(
-		filePath: string, linkPath: string, linkText: string, routesVariable = ROUTES_VARIABLE, parentRoutePath?: string) {
+		filePath: string,
+		linkPath: string,
+		linkText: string,
+		routesVariable = DEFAULT_ROUTES_VARIABLE,
+		parentRoutePath?: string
+	) {
 		let className: string;
 		const fileSource = TsUtils.getFileSource(filePath);
 		const relativePath: string = Util.relativePath(this.targetPath, filePath, true, true);
-		routesVariable = routesVariable || ROUTES_VARIABLE;
+		routesVariable = routesVariable || DEFAULT_ROUTES_VARIABLE;
 		className = TsUtils.getClassName(fileSource.getChildren());
 		this.requestImport([className], relativePath);
 
@@ -131,6 +136,9 @@ export class TypeScriptFileUpdate {
 							let childrenProperty: ts.PropertyAssignment = syntaxList
 								.getChildren().filter(filterForChildren)[0] as ts.PropertyAssignment;
 							let childrenArray: ts.ArrayLiteralExpression = null;
+
+							// if the target parent route already has child routes - get them
+							// if not - create an empty 'chuldren' array
 							if (childrenProperty) {
 								childrenArray = childrenProperty.getChildren()
 									.filter(element => element.kind === ts.SyntaxKind.ArrayLiteralExpression)[0] as ts.ArrayLiteralExpression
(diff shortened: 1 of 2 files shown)


packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -5,6 +5,8 @@ import { FS_TOKEN, IFileSystem } from "../types/FileSystem";
 import { Util } from "../util/Util";
 import { TypeScriptUtils as TsUtils } from "./TypeScriptUtils";
 
+const ROUTES_VARIABLE = "routes";
```
ViktorSlavov, line 8: Should this be exported?

packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -65,64 +66,120 @@ export class TypeScriptFileUpdate {
...
+		linkPath: string,
+		linkText: string,
+		routesVariable = DEFAULT_ROUTES_VARIABLE,
+		parentRoutePath?: string
+	) {
 		let className: string;
 		const fileSource = TsUtils.getFileSource(filePath);
 		const relativePath: string = Util.relativePath(this.targetPath, filePath, true, true);
-
+		routesVariable = routesVariable || DEFAULT_ROUTES_VARIABLE;
```
damyanpetev, line 81: There's literally no need to do this, that's the who point of default-initialized parameters - they assign the default value if none is passed. To skip over them you provide `undefined` to get the default. The proper call in the tests should be
`tsUpdate.addRoute("path/to/component", "child-1", "child one", undefined, "parent");` instead of `tsUpdate.addRoute("path/to/component", "child-1", "child one", "", "parent");`

That leads me to the two optional params at the end - frankly if you find the jumping over `routes` too annoying (it was really meant to be an the last one) - you can swap them and default the parent route to null as I think the route variable override is used even less, if at all.
Alternatively, can be split into two public facing methods that map back to this one (which would be hidden)

packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -65,64 +66,120 @@ export class TypeScriptFileUpdate {
...
-
-						return ts.updateArrayLiteral(array, elements);
-					} else {
-						return ts.visitEachChild(node, visitRoutesVariable, context);
-					}
+					};
+				} else {
+					conditionalVisitor = (node: ts.Node): ts.Node => {
+						if (node.kind === ts.SyntaxKind.ObjectLiteralExpression) {
+							if (!node.getText().includes(parentRoutePath)) {
```
damyanpetev, line 121: That is really not a very robust way to check if that's the parent route you are looking for - the text of the whole object can contain false matches. You should be able to access `properties` on the node and look for the one with name 'path' I guess and get it's value (hint: will require interface cast to get to property assignment I think)

packages/core/typescript/TypeScriptFileUpdate.ts
damyanpetev: For more on the params see https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters
Also if curious, check out this [playground](https://www.typescriptlang.org/play/index.html#code/GYVwdgxgLglg9mABAQwCaoPIBtUGUIAWccWAFAB4BciYIAtgEYCmATgDSICeiAvIgIwcAXr0QAmAJSIA3gFgAUIiWIICAM4kmAOixwA5qQAGVRABJp5AL6GJAbgXKV6zTv1HO1c52t2Hy1WAaWNq6BoZCntJCPvaKyixMUCAsSOSIANRcGYhCsZYKCgFBIW5omDj4RCSkAMwc4KhMwDBgTKiIbDUSvvJAA) and change the target to ES5 to see what the functionality polyfil does:
```js
// function addOldSchool(x, y = 1, z = 2) {
function addOldSchool(x, y, z) {
    if (y === void 0) { y = 1; }
    if (z === void 0) { z = 2; }
```

packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -65,64 +66,120 @@ export class TypeScriptFileUpdate {
...
-						const errorRouteNode = nodes.filter(element => element.getText().includes(notFoundWildCard))[0];
-						let resultNodes = null;
-
-						if (errorRouteNode) {
-							resultNodes = nodes
-								.slice(0, nodes.indexOf(errorRouteNode))
-								.concat(newObject)
-								.concat(errorRouteNode);
+				let conditionalVisitor: NodeVisitor;
+				if (parentRoutePath === undefined) {
```
damyanpetev, line 89: This might be a nit-pick, but there's something about those big chunks inside the if-else. This can just as easily be two visitor variables and an inline condition to pick one before creating the main visitor.

packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -5,6 +5,8 @@ import { FS_TOKEN, IFileSystem } from "../types/FileSystem";
 import { Util } from "../util/Util";
 import { TypeScriptUtils as TsUtils } from "./TypeScriptUtils";
 
+const DEFAULT_ROUTES_VARIABLE = "routes";
+declare type NodeVisitor = (node: ts.Node) => ts.Node;
```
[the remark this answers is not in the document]
ViktorSlavov: 🤦

PR #811, opened 2020-11-14
fix(*): ssr breaks because of localStorage ref

kdinev:

The global reference to `localStorage` in the fake backend is causing SSR to break.


Lipata:
https://github.com/IgniteUI/igniteui-cli/pull/812

Issue #847, opened 2021-06-03
Enable strict templates for angular templates


Lipata:

Lipata:
Most of the angular templates are without errors when strict templates are enabled. The application scenarios are still having errors, but first, we should do that in the Ignite UI for the Angular samples browser and then put the changes in the CLI templates as well.

PR #848, opened 2021-06-08
fix(angular-templates): ensure correct templates for strict compiler modes

ViktorSlavov:
Related #847.

Fixed template error that might show up when `angularCompilerOptions.strictTemplates` is enabled.
Fixed **some** .ts files that might cause errors when `compilerOptions.strict` is enabled.

Pending:
Fix scenario templates for fintech(tree) / awesome-grid, as they will first be striclty typed in the samples.


IvayloG (approved):

packages/igx-templates/igx-ts/autocomplete/autocomplete-extended/files/src/app/__path__/__filePrefix__.component.ts
```diff
@@ -1,28 +1,35 @@
...
 
-    public getPostalCode(event) {
-        const targetRegion = this.extendedTowns.regions.filter((r) => r.name === event.newSelection.group.label)[0];
-        this.postalCode = targetRegion.towns.filter(t => t.name === event.newSelection.value)[0].postalCode;
+    public getPostalCode(e: ISelectionEventArgs) {
+
+        // Flatten the structure of the regions array
+        const arr = this.regions;
+        const allTowns = arr.map(region =>
+          region.towns.map(town => ({name: town.name, postalCode: town.postalCode}))).flat();
```
damyanpetev, line 28: Is `{name: town.name, postalCode: town.postalCode}` not the same as just the town object? And if so this is really `town => town` so no transform at all, and the same can be accomplished with 
```ts
townsExtended.map(region => region.towns).flat()
```
Also no point in the `arr` variable or the comment above it if the above suggestion is applies since it's fairly obvious what it does :)

IvayloG (approved):

Lipata (approved):

PR #865, opened 2021-07-30
fix(igx-ts): remove deprecated properties from scenario templates

ViktorSlavov:
Some icons were using the deprecated 'color' property
Rename referance from `cell.row.rowData` => `cell.row.data`


Lipata (approved):

dafo (changes requested):
The "color" property is still used in fintech-grid and fintech-tree-grid templates

dafo (approved):

PR #931, opened 2021-11-09
Removing unused assests in the list component

valadzhov:
The unused asset was causing an error when adding more than one list component:
image


valadzhov:
The unused asset was causing an error when adding more than one list component:
image

PR #932, opened 2021-11-09
Webcomponents project improvements

dafo:
- Removing assets for List template
- Styling NavDrawer header item
- Removing description for all templates 
- Renaming "Grids & Lists" group to "Lists" as grids are still not included
- Fixing some minor issues


Lipata (approved):

Lipata:
Only this is left, but not necessary to be fixed, knowing it is in the core package:

image

PR #942, opened 2021-11-18
fix: add igx-scrollbar class to body, #941

### LINKED ISSUE #941: igx-scrollbar class is missing at root app level upon export
### Description  
igx-scrollbar class is missing from body of generated project.

- Ignite UI CLI version: 9.0.0
- Framework: Angular
- Project type: igx-ts

### Steps to reproduce  

1. Open AB
2. Add a grid
3. Change to Fluent dark theme
4. See in App preview that all Grid scrollbars are dark
5. Export the app
6. Run it
7. All Grid Scrollbars are light

### Result  
Scrollbars of generated grids should be dark.

### Expected result  
Scrollbars of the generated grids are light.

wnvko:
Closes #941   

Added  `igx-scrollbar` in body classes.
Also added migration for adding ` igx-scrollbar` class when updating to 9.0.0.


packages/cli/migrations/migration-collection.json
```diff
@@ -30,6 +30,11 @@
             "version": "5.0.3",
             "description": "Updates Ignite UI for Angular project from CLI 5.0.x to 5.0.3",
             "factory": "./update-5_0_3"
+        },
+        "migration-07": {
```
damyanpetev, line 34: Hrm, the situation w/ migrations in the CLI is a bit odd. Keep in mind this will only run for `ng update igniteui-cli` which is not necessarily present in all projects - the rest should use the schematics package only, which is the preferred method ATM anyway.

Not sure, but I believe the schematics should be grouped with the CLI main package for migrations (check package.json) so you should be able to write a migration there that can apply in both cases.

damyanpetev: On second thought, not sure how wise it is to modify existing projects' scrollbars. While I think it'd be an improvement, it could be undesirable for some and it's always iffy to do that on existing project

packages/cli/migrations/update-9_0_0/index.ts
```diff
@@ -0,0 +1,10 @@
+// tslint:disable:no-implicit-dependencies
+import { Rule, SchematicContext, Tree } from "@angular-devkit/schematics";
+import { addTypography } from "@igniteui/cli-core";
+
+export default function(): Rule {
+	return (host: Tree, context: SchematicContext) => {
+		context.logger.info("Updating project to Ignite UI CLI 9.0.0");
+		addTypography(host, "igx-scrollbar");
```
damyanpetev, line 8: so this should be renamed to `addClassToBody` then?

damyanpetev (commented):
Also, equivalent scrollbar change in Web Compoents?

packages/cli/migrations/migration-collection.json
wnvko: migration was reverted now

Lipata (approved):

PR #954, opened 2021-11-23
Update 'build' command to bundle for production

### LINKED ISSUE #943: IgcDataGrid brakes all igc-ripple elements on the page
### Description  
When you add an IgcDataGrid view, all other igc-elements on the page stop working properly. The NavDrawer items take focus but no navigation is executed. One should click on the item's text to navigate to the corresponding view.

- Ignite UI CLI version: 9.0.0-rc.0
- Framework: Web components

### Steps to reproduce  

1. Create a new Web components project
2. Add a Grid view
3. Run he project
4. Click the "Grid" item in the navigation but not on the text

### Result  
The navigation item takes focus but no navigation is executed

### Expected result  
You should be navigated to the Grid view

### LINKED ISSUE #951: Implement bundling for Webcomponents projects
### Description  
use a bundler to treeshake the generated project for production. Consider Rollup and Webpack, for example.

- Ignite UI CLI version: x.x.x
- Framework: Web Components
- Project type: igc-ts

### Expected result  
The generated project is bundled for deploying on production.

### LINKED ISSUE #956: Add more views and scenarios to Web components project
### Description  
More DataGrid/Gauges/Charts views should be added. We can use the existing views in Ignite UI for Angular project.
We should add some complex scenarios.

- Framework: Web Components

### Expected result  
**Grid templates:**

- [x] Grid summaries -> [Sample](https://staging.infragistics.com/products/ignite-ui-web-components/web-components/components/grids/data-grid-column-summaries)
- [x] Grid editing -> [Sample](https://staging.infragistics.com/products/ignite-ui-web-components/web-components/components/grids/data-grid-cell-editing)

**Gauges:**

- [x] Linear gauge -> [Sample](https://staging.infragistics.com/products/ignite-ui-web-components/web-components/components/linear-gauge)
- [x] Radial gauge -> [Sample](https://staging.infragistics.com/products/ignite-ui-web-components/web-components/components/radial-gauge)

**Charts:**

- [x] Financial chart -> [Sample](https://staging.infragistics.com/products/ignite-ui-web-components/web-components/components/charts/types/stock-chart)

**Scenarios:**

- [x] Form view -> [Sample](https://staging.infragistics.com/products/ignite-ui-web-components/web-components/components/inputs/form)

dafo:
Closes #951 
Closes #943 
Closes #956 

This should not be merged before investigating an issue when a template imports anything from 'igniteui-webcomponents'

Example error:

ERROR in ./node_modules/igniteui-webcomponents/index.js 1:2377-2456
Module not found: Error: Can't resolve './components/icon/icon.registry' in 'C:\work\github\igniteui-cli\output\IG Project 12\node_modules\igniteui-webcomponents'
Did you mean 'icon.registry.js'?
BREAKING CHANGE: The request './components/icon/icon.registry' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ./src/app/badge/badge.ts 1:0-71 1:71-87 1:88-105
 @ ./src/index.ts 1:69-98


dafo:
This is fixed.
More information on that error -> [here](https://github.com/webpack/webpack/issues/11467#issuecomment-691873586)

Lipata (dismissed):

gmurray81 (changes requested):
Seems like that'll do the trick. Personally, I do like there to be a legacy mode so that you can build for older browser like in my example, but I think we were being told we didn't need to support IE111 any longer, and this certainly makes for a simpler webpack config. You might not technically need to force lit-html through babel if not supporting older browsers. The main reason I was doing that was because lit-html ONLY distributes esm modules, so in order to make them usable in ie11, they needed to be fed through babel. Similarly, you may not need to feed our stuff through babel either. You may be able to change the exclusion so that it just excludes all of node_modules.

Unless you want to try to add the legacy support, I'd see if it works fine by just excluding all of node_modules in the exclude statement

packages/cli/templates/webcomponents/igc-ts/projects/empty/files/webpack.config.mjs
```diff
@@ -0,0 +1,72 @@
...
+            }
+          }],
+          "@babel/preset-typescript"
+        ],
+        "plugins": [
+          "@babel/plugin-proposal-class-properties",
+          "@babel/plugin-transform-runtime"
+        ]
+      },
+      exclude:
```
gmurray81, line 56: This is the portion I'm referring to where maybe just change this exclude to exclude all of node_modules and see if it still behaves appropriately, if only trying to support browsers that needs es22015 and upward.

packages/cli/templates/webcomponents/igc-ts/projects/empty/files/webpack.config.mjs
```diff
@@ -0,0 +1,72 @@
...
+        function(modulePath) {
+          return /node_modules/.test(modulePath) &&
+            !/igniteui-webcomponents/.test(modulePath) &&
+            !/lit-html/.test(modulePath);
+        }
```
gmurray81, line 61: This is the portion I'm referring to where maybe just change this exclude to exclude all of node_modules and see if it still behaves appropriately, if only trying to support browsers that needs es22015 and upward.

gmurray81: The original intent of that exclusion was to make it run our stuff through babel and lit-html through babel so that they would be downleveled approrpriately for older browsers. If you aren't trying to do that with this generated app though, then things will be faster if babel skips processing our stuff.

Unless of course this causes webpack not to tree shake our stuff for some reason... but I don't think that will be the case.

gmurray81: Ah, one exception to this though, babel does another couple neat things whereby it will do the downleveling, and can also automatically add required polyfills for certain browsers.

So this may actually depend on what "level" the code is that is in igniteui-webcomponents, and wether it needs any more polyfilling or downleveling to target all the browsers we need to target? If so, it may be safer to keep this as is and feed everything through babel. it just will be a little slower that way, potentially.

gmurray81: Its possible it will do some processing of the transitive dependencies anyhow these days... Not sure... I just remember needing to add this statement back when I put this config together for it to actually consider downleveling lit-html for IE11, and additionally our stuff (although we could have just pointed at the es5 entry points alternatively). Anyhow, I'd experiment with whether you can use the simpler exclusion, unless you suspect you would benefit from the auto poly-filling.

dafo: Thank you for your comments, @gmurray81 
I left only the node_modules in the exclude statement and it looks like the project is bundled fine (this will be additionally tested before closing the PR). I asked @gedinakova to have a look at this case and it looks like igniteui-webcomponents do not support IE11, so as to my understanding we will not have any issues with that.
Can you please check the last commit and tell if this is what you meant in your previous comments.

gmurray81 (approved):

Lipata (approved):

PR #963, opened 2021-12-09
Updating path to bootstrap light theme

dafo:
The path to the themes is updated to support dark themes in addition to the light ones.


Lipata:
Do not merge until new version of the web-components is releases and in contains that change

Lipata (approved):

Issue #969, opened 2022-01-12
Views are clipped by navigation in Web Components projects


damyanpetev:
### Description  
Some views are clipped under the side navigation in in Web Components projects. Currently saw the Home and Calendar view, but others could be affected as well.

- Ignite UI CLI version: 9.0.4
- Framework: Web Components

### Steps to reproduce  

1. Run the Step-By-Step mode and create a new Web Components project
2. Add a Calendar view
3. Run the project
4. Observe the page layout
5. Navigate to the Calendar view and observe the layout again

### Result  
Views appear partially under the nav drawer:
![image](https://user-images.githubusercontent.com/3198469/149154837-f1146d85-aa5c-4781-ac60-fc4cd9b3e515.png)
![image](https://user-images.githubusercontent.com/3198469/149154896-b5db1de7-06b3-485d-b00c-e47ff2492801.png)

### Expected result  
If the side navigation is static, views should not be obscured by it and be fully visible to the side like the Grid Editing template is

PR #979, opened 2022-01-14
Adding Empty view for WC project

### LINKED ISSUE #967: Separate empty project template for Web Components
## Is your feature request related to a problem? Please describe.
Currently there is only one project template available for Web Components (which is actually indicated as empty) that contains a Nav Drawer side navigation pre-configured.

## Describe the solution you'd like
The Ignite UI CLI should provide an actual empty project template in case the side-navigation is not desirable for the further development on the project.

dafo:
Closes #967  

Adding Empty view with no side navigation in it. The previously existing "empty project" is now named "Default Side Navigation"


dafo:
This also resolves https://github.com/IgniteUI/igniteui-cli/issues/969

Please, check all templates, specifically "Card" and "Navbar" as they needed specific updates.

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/package.json
```diff
@@ -0,0 +1,85 @@
+{
```
damyanpetev, line 1: @dafo No need to copy the `pacakge.json` (and the 7 other config files in the root alone, haven't checked folders) that are identical between projects, it'd be very cumbersome to maintain multiples.
Move all the shared files in a base (see Angular's setup for working example), put different/additional files in their respective projects.

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/index.ts
```diff
@@ -0,0 +1,44 @@
+import { ControlExtraConfiguration, defaultDelimiters, ProjectTemplate, Util } from "@igniteui/cli-core";
```
damyanpetev, line 1: Same logic applies to the Project logic itself, which is again (sans a default title) the same - move to a base project and inherit the shared logic

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/package.json
damyanpetev: On that note, the current setup looks off to me:
- Index.html imports two modules - index and app. Would be nice to have a single entry point, index (or main) may very well import App internally, right?
- The empty project not having an app file is weird, it's index.html still using `<app-navigation></app-navigation>` in the markup even more so (and it's wrong as it stands, careful!). Like there's still an app as a concept at least. The only difference is the empty proj has basically an outlet being the entire layout.

damyanpetev: To the last point, the router outlet being external to the app doesn't look quite right, it'll definitely mess with creating more complicated layouts. Is that a limitation of some sorts for the router in use?

damyanpetev (changes requested):

packages/cli/templates/webcomponents/igc-ts/card/default/files/src/app/__path__/__filePrefix__.ts
```diff
@@ -20,7 +20,7 @@ export default class $(ClassName) extends HTMLElement {
     this.innerHTML = `
```
damyanpetev, line 20: Also, general: So the default Home component is `home.component` but the rest of our templates are simply `card` and so on? That's inconsistent at least and then there's Lint:
image

Not that I'm saying `home.component` passing the check is okay (it's _not_), but first perhaps our components should be consistent and then they shouldn't break the lint rules

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/index.html
```diff
@@ -17,7 +17,7 @@
       padding: 25px;
```
damyanpetev, line 17: `:host` ?! What would that be in this setting?

damyanpetev: Also, could we not use an actual css file for this since it's not a component, but the index.html file?

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/index.html
```diff
@@ -17,7 +17,7 @@
...
     }
 
-    #outlet {
+    app-root {
```
damyanpetev, line 20: Also also, is that supposed to be part of the app component itself?

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/src/app/app.ts
```diff
@@ -0,0 +1,44 @@
...
+  IgcNavDrawerComponent,
+  IgcNavDrawerItemComponent,
+  IgcNavDrawerHeaderItemComponent,
+  IgcRippleComponent,
+);
+
+export class App extends HTMLElement {
+  connectedCallback() {
+    this.innerHTML = `
+      <div class="outer-wrapper">
```
damyanpetev, line 22: The outer wrapper doesn't seem to do much, not entirely sure it's needed

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/src/index.ts
```diff
@@ -0,0 +1,2 @@
+import './app/app';
```
damyanpetev, line 1: That's identical to the file in _base, remove

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/index.ts
```diff
@@ -0,0 +1,18 @@
+import { ProjectTemplate } from "@igniteui/cli-core";
+import * as path from "path";
+import { BaseIgcProject } from "../_base";
+
+export class SideNavProject extends BaseIgcProject implements ProjectTemplate {
+	public id: string = "webcomponent";
```
damyanpetev, line 6: id "webcomponent"? That's affects the CLI parameters of `ig new`, pretty sure it should be `side-nav` for example. Also should be reflected in the docs/wiki

damyanpetev (changes requested):
Can't add a comment to the package.json file since it's just a rename, but the CLI version embedded in the project is static - i.e. 
```json
"igniteui-cli": "9.0.5-beta.1",
```
That's actually a good thing ATM, since we are making changes to the project structure, but in general it should be at least `~`. Thing is we don't really have an upgrade path, which can cause some trouble, something to consider

packages/cli/templates/webcomponents/igc-ts/projects/_base/files/index.html
```diff
@@ -0,0 +1,34 @@
+<!doctype html>
+<html lang="en-GB">
+
+<head>
+  <meta charset="utf-8">
+  <base href="/">
+  <link rel="stylesheet" href="./node_modules/igniteui-webcomponents/themes/bootstrap.css">
```
damyanpetev, line 7: The `index.html` in the empty proj is using themes/**light**/bootstrap.css, I'm assuming that's updated from the new package. This one should be too, right? Otherwise it won't work

damyanpetev:
Oh also:
image

packages/cli/templates/webcomponents/igc-ts/card/default/files/src/app/__path__/__filePrefix__.ts
dafo: home.component.ts and not-found.component.ts are renamed to home.ts and not-found.ts, respectively.
I logged issue #997 for the missing .js extensions.

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/src/app/app.ts
dafo: Yeah. I removed it and changed some stylings in that file that affect charts and gauges,

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/index.ts
dafo: I changed it and updated the wiki

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/index.html
dafo: No, we need to apply those styles to the <app-root> element in index.html, too.

packages/cli/templates/webcomponents/igc-ts/grid/default/files/src/app/__path__/__filePrefix__.ts
```diff
@@ -45,9 +45,8 @@ export default class $(ClassName) extends HTMLElement {
       :host {
         height: 100%;
         margin: 0px;
-        padding-left: 275px;
         padding-right: 20px;
-        width: calc(100% - 275px);
+        width: calc(100% - 600px);
```
damyanpetev, line 49: Woah that looks weird. I guess it can be looked at separately though

packages/cli/templates/webcomponents/igc-ts/projects/_base/files/index.html
```diff
@@ -0,0 +1,17 @@
+<!doctype html>
+<html lang="en-GB">
+
+<head>
+  <meta charset="utf-8">
+  <base href="/">
+  <title>Ignite UI for Web Components</title>
```
damyanpetev, line 7: Fairly certain the title is supposed to be templated from the user-provided one, no?

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/index.html
```diff
@@ -0,0 +1,18 @@
+<!doctype html>
+<html lang="en-GB">
```
damyanpetev, line 2: This is now identical to the base, so it's not needed as an override

packages/cli/templates/webcomponents/igc-ts/projects/side-nav/files/styles.css
```diff
@@ -0,0 +1,5 @@
+html, body {
```
damyanpetev, line 1: This is also identical to base now

packages/cli/templates/webcomponents/igc-ts/projects/_base/files/index.html
dafo: I was using the Ignite UI for Angular template as a guide so I think it is fine the way it is now?

packages/cli/templates/webcomponents/igc-ts/grid/default/files/src/app/__path__/__filePrefix__.ts
dafo: Removing the styles in styles.css and app.ts broke the grid views. That's why I updated the width here, No other views were affected. However, you are right this does not look nice. I will look into the grid issue further,

packages/cli/templates/webcomponents/igc-ts/projects/_base/files/index.html
damyanpetev: Hah, could very well be wrong in Angular's templates too :) 
I guess it's a preference, though kind of odd you give a name for the app and that's not in the title, right? Can be left for a separate item as well

damyanpetev (approved):
There are some more things to improve (layout styling, build) and some cleanup, but will leave those for other tasks. The overall project structure is a lot cleaner now in addition to having a more basic project with empty layout. So, LGTM

PR #1001, opened 2022-02-11
Web Components templates use LitElement

### LINKED ISSUE #989: Research LIT without shadow DOM for web-components project
Anything other than the default theme, cannot be applied when we have LIT component with shadow DOM (this will change when https://github.com/IgniteUI/igniteui-webcomponents/issues/228 is implemented). We want to use LIT in the CLI templates and then in the CodeGen in order to use LIT functionalities.

dafo:
Closes #989  .  


dafo:
Add events in renderer where available.
Check if we can use firstUpdated instead of updated where necessary.

jackofdiamond5:
It seems like the name of the project was not placed in the package.json:
image

jackofdiamond5:
Also, there seem to be quite a few linting errors in multiple files in the generated project, namely using single quotes instead of double quotes, tabs instead of spaces, trailing commas and using `&gt;` instead of `>`. I'm using `eslint v2.2.2`.

jackofdiamond5:
We may want to make sure the whole tsconfig file is properly formatted:
image

jackofdiamond5:
I think currently the project structure kind of looks like an "Angularized" version of WC. If we're talking about a components library (which I think is what our generated app is supposed to be) and if you generate a new one with `npm init @open-wc`, there seem to be a few differences between what is generated by them and us. For example, the `assets` folder is not in `src`. Also, all spec files are separated in a different directory and it's not in the same directory as the components:
image

In our case they are in the same directory and this is what makes it look very Angular-ish to me. The redundant `app` directory can also be removed in my opinion and just dump all of the ts files directly into `src`, since that's what open-wc seem to be doing. I think we should try to keep close to whatever open-wc generate so that it makes sense to open-wc developers.

Lipata (approved):

Issue #983, opened 2022-01-24
Web components projects should use the general TypeScriptFileUpdate.ts 


dafo:
### Description  
Currently web components projects use own TypeScriptFileUpdate.ts file, located inside their own folder. It duplicates much of the logic inside the "general" file placed under "packages/core/typescript". 
All projects must use the general file. It should be updated to support generation of project different than Angular.
N. B. The logic should be platform agnostic. Where we check for Ng stuff - we should set it as parameters instead of looking what platform the generated project is.

We should refactor the `addRoute` method. This method is actually doing two things, at least. It adds routes as well as it adds imports. We should try to split this into two separate methods. Note, to do this we may need to create separate visitor for imports and routes. - https://github.com/IgniteUI/igniteui-cli/pull/1060#discussion_r988903737

- Ignite UI CLI version: 9.x.x
- Framework: Web Components
- Project type: igc-ts

dafo:
https://github.com/IgniteUI/igniteui-cli/pull/1060#discussion_r988903737

PR #1060, opened 2022-09-19
Support for nested routing modules

dafo:
Support nested routing modules.
Support importing exported classes.


packages/cli/templates/webcomponents/TypeScriptFileUpdate.ts
```diff
@@ -1,4 +1,4 @@
-import { App, FS_TOKEN, IFileSystem, TypeScriptUtils, Util } from "@igniteui/cli-core";
```
wnvko, line 1: in the end entire `TypeScriptFileUpdate` is mostly the same class exported in `cli.core`. I believe this was added here when POC was created. It is now time to remove this class and use the one from `core`. The changes added in this class should be transferred to the one in the `core`.

packages/cli/templates/webcomponents/TypeScriptFileUpdate.ts
```diff
@@ -33,8 +33,15 @@ export class TypeScriptFileUpdate {
 		this.initState();
 	}
 
-	public addRoute(filePath: string, linkPath: string, linkText: string, routesVariable = DEFAULT_ROUTES_VARIABLE) {
-		this.addRouteModuleEntry(filePath, linkPath, linkText, routesVariable);
+	public addRoute(
```
wnvko, line 36: This method is actually doing two things, at least. It adds routes as well as it adds imports. We should try to split this into two separate methods. Note, to do this we may need to create separate visitor for imports and routes, and all this should be done in `core` `TypeScriptFileUpdate`.

packages/cli/templates/webcomponents/TypeScriptFileUpdate.ts
dafo: There is an issue about that - https://github.com/IgniteUI/igniteui-cli/issues/983

packages/cli/templates/webcomponents/TypeScriptFileUpdate.ts
dafo: I added this to the requirements in issue https://github.com/IgniteUI/igniteui-cli/issues/983

PR #1063, opened 2022-09-19
feat(wc): add code gen related changes

Lipata:


PlamenaMiteva:
Seems that the routing of the generated apps is broken, since it generates the project but fails to start it

PR #1070, opened 2022-10-17
Provide default options when adding a template

dafo:


packages/cli/lib/commands/add.ts
```diff
@@ -114,6 +114,13 @@ command = {
 		}
 	},
 	async addTemplate(fileName: string, template: Template, options?: AddTemplateArgs): Promise<boolean> {
+		if (!options) {
+			options = new Object();
```
wnvko, line 118: you can go in more js way like this:
```
options = {
  parentName": "app",
  parentRoutingModulePath": "src/app/app-routing.ts",
  selector": "app-" + template.id
}
```

wnvko (dismissed):

wnvko (dismissed):

wnvko (approved):

Lipata (approved):

PR #1096, opened 2023-01-26
Dpetev/yargs bump 10.1.x

damyanpetev:
Closes # .  

### DIFF
--- a/packages/cli/lib/cli.ts
+++ b/packages/cli/lib/cli.ts
@@ -53,6 +53,12 @@ export async function run(args = null) {
 	.command(test as any)
 	.command(list as any)
 	.command(upgrade as any)
+	// TODO: Please do one of the following:
+	// - Disable version with `yargs.version(false)` if using "version" as an option
+	// - Use the built-in `yargs.version` method instead (if applicable)
+	// - Use a different option key
+	// https://yargs.js.org/docs/#api-reference-version
+	.version(false)
 	.options({
 		version: {
 			alias: "v",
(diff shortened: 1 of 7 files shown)


damyanpetev:
@PlamenaMiteva @Lipata 
The changes in the PR are for the most part bumping yargs to latest, though I'm assuming there are yarn.lock diffs in latest and it might not be feasible to reuse this PR and might need to be recreated. If not, this could also be retargeted just as easily.

If useful here are the steps and some issues I found along the way that need to be resolved.
```bash
yarn workspace igniteui-cli add yargs@latest
```
And since we didn't have those, I figured types would be nice and help with the migration process, so also
```bash
yarn workspace igniteui-cli add @types/yargs@latest --dev
```
And they sort of did, the notable change being that `.parseSync()` is now needed to go back to the old behavior we used. However, there's an issue with the options we pass to `.command()`, because our objects contain extra stuff and thus the multiple `as any` type asserts in the PR. I believe those confuse the overload matching and it ends up going for a different version, though the type errors might be perfectly valid too - needs to be checked against typing and docs and adjusted. 
_PS: If we update TypeScript [`satisfies`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) looks nice :)_

One more thing I landed on - the `typescript-json-schema` we use for the config schema failed trying to parse the new yargs type (which is why I reverted those to see if the build will work). Tried updating, though that didn't help - might be reliant on the typescript version itself so update there might be needed. There's also the question of why was the schema trying to parse the yargs types in the first place - there's a good chance it's better to run it with its own config which is possible looking at the docs, cuz I'm pretty sure it doesn't need to use those types to produce the schema for a few of our interfaces.

Lastly, not sure if it's just on my end, but I can't run tests locally due to lint fails which mysteriously don't happen on the build machines. @Lipata this might be a separate task/issue, but there's something wrong with the lint setup, ignoring the fact that tslint should be replaced with eslint too.

PR #1104, opened 2023-03-20
Fix extensions in routes in webcomponents templates

onlyexeption:


Lipata:
After some research, according to [official Typescript documentation](https://www.typescriptlang.org/docs/handbook/esm-node.html), imports with .js extensions should be used. Closing the PR.

PR #1127, opened 2023-05-18
chore(igx): update empty projects with material icons extended

igdmdimitrov:
Update Angular and Web Components projects with material icons extended npm package and add CSS sprite to angular.json


packages/igx-templates/igx-ts/projects/_base/files/package.json
```diff
@@ -19,6 +19,7 @@
     "@angular/platform-browser": "~16.0.0",
     "@angular/platform-browser-dynamic": "~16.0.0",
     "@angular/router": "~16.0.0",
+    "@igniteui/material-icons-extended": "^3.0.0",
```
damyanpetev, line 22: Yeah, can we not blindly hardcode the package in the project? I'd much prefer if the package and angular.json modification are done dynamically when there's actual usage (for example a view that has material extended icons).

Issue #1030, opened 2022-06-02
Angular project includes redundant packages


damyanpetev:
### Description  
A number of packages listed as dev dependencies in the Angular project's package.json are unused - such as `codelyzer` which was dropped in the switch to eslint I believe, `protractor` and `ts-node` which are for e2e which is no longer in the default project. There seem to be even a few more.

- Ignite UI CLI version: 9.1.x
- Framework: Angular
- Project type: igx-ts

### Steps to reproduce  

1. Create a new Angular project

### Result  
package.json:
```json
  "devDependencies": {
    "@angular-devkit/build-angular": "~13.3.0",
    "@angular-eslint/builder": "13.1.0",
    "@angular-eslint/eslint-plugin": "13.1.0",
    "@angular-eslint/eslint-plugin-template": "13.1.0",
    "@angular-eslint/schematics": "13.1.0",
    "@angular-eslint/template-parser": "13.1.0",
    "@angular/cli": "~13.3.0",
    "@angular/compiler-cli": "~13.3.0",
    "@types/jasmine": "~3.10.0",
    "@types/node": "^12.11.1",
    "@typescript-eslint/eslint-plugin": "5.3.0",
    "@typescript-eslint/parser": "5.3.0",
    "codelyzer": "^6.0.0",
    "eslint": "^8.1.0",
    "igniteui-cli": "~9.1.1",
    "jasmine-core": "~4.0.0",
    "jasmine-spec-reporter": "~7.0.0",
    "karma": "~6.3.0",
    "karma-chrome-launcher": "~3.1.0",
    "karma-coverage": "~2.1.0",
    "karma-jasmine": "~4.0.0",
    "karma-jasmine-html-reporter": "~1.7.0",
    "protractor": "~7.0.0",
    "ts-node": "~9.1.1",
    "typescript": "~4.5.2"
  }
  ```

### Expected result  
Besides the additional `igniteui-cli` or `@igniteui/angular-schematics` dev dependency we add, there shouldn't be extras unless we do use them ( guessing eslint is an addon) and should be closer to a clean ng cli project:
```json
  "devDependencies": {
    "@angular-devkit/build-angular": "~13.2.2",
    "@angular/cli": "~13.2.2",
    "@angular/compiler-cli": "~13.2.0",
    "@types/jasmine": "~3.10.0",
    "@types/node": "^12.11.1",
    "jasmine-core": "~4.0.0",
    "karma": "~6.3.0",
[...]

PR #1134, opened 2023-05-26
chore(igx-ts): remove igniteui-theming package

Lipata:
#1030


Lipata:
The problem is resolved in the theming package, so again the dependency is not needed: https://github.com/IgniteUI/igniteui-theming/releases/tag/v1.4.5

PR #1146, opened 2023-09-13
feat(igr-ts): add igr-ts proj type

### LINKED ISSUE #1147: Switch the React 18 project to use the new grid
We need to generate the grid templates using the new grid for the new React 18 project.

### LINKED ISSUE #1153: React app not running on the default port
### Description  
When you use the CLI to bootstrap a React project, the app runs on Vite's default port (5173), not the one specified in the CLI's last step (3003).

- Ignite UI CLI version: 12.0.6-beta.0
- Framework: React

### Steps to reproduce  

1. Run the cli, and enter some project name
2. Selet React for the framework
3. Choose "Ignite UI for React TS"
4. Use the "Default top navigation"
5. Complete and run
6. Choose the default port (3003)

### Result  
The application starts at the default port for Vite: 5173

### Expected result  
Port should be 3003

### LINKED ISSUE #1158: No need for react-native-get-random-values lib
### Description  
When you bootstrap a project with the CLI, there is this dependency: ```react-native-get-random-values```. It's not being used anywhere, so it might as well be removed.

- Ignite UI CLI version: 12.1.0-beta.1
- Framework: React
- Project type: igr-ts

### Steps to reproduce  

1. Run the CLI and configure a React with TS project

### Result  
Open the package.json and notice that there is the ```react-native-get-random-values``` dependency.

### Expected result  
The ```react-native-get-random-values``` should not be in the package.json and not installed.

### LINKED ISSUE #1159: Clean up all files that are not needed in the React projects
such as: .env and the editorconfig

onlyexeption:
Closes # .
Closes #1147 
Closes #1153 
Closes #1158 
Closes #1159


packages/cli/templates/react/igr-ts-es6/projects/_base/files/src/serviceWorker.tsx
igdmdimitrov, line 1: @damyanpetev , do we need this Service Worker? Is it something that was needed for Create React Apps?

damyanpetev:
@onlyexeption Assuming the existing Ignite UI for React project was `igr-es6` can we name this one just `igr-ts` to follow the pattern? :)

packages/cli/templates/react/igr-ts-es6/projects/_base/files/src/serviceWorker.tsx
damyanpetev: Honestly, I don't think we need more than the template we're basing this on initially.
No idea what that file is doing here. I don't see it in the vite starter template, so I can assume the initial project structure is copied from the old one? If so, might be more extra files, you might want to do a full comparison :)

packages/cli/templates/react/igr-ts/projects/top-nav/files/src/app/App.css
```diff
@@ -0,0 +1,59 @@
...
+	box-sizing: border-box;
+	border-bottom: 1px solid transparent;
+  }
+  nav ul li.active {
+	border-bottom: 2px solid #fff;
+  }
+  nav ul li:not(.active):hover {
+	border-bottom: 1px solid #fff;
+  }
+
```
jackofdiamond5, line 59:
```suggestion

```

packages/cli/templates/react/igr-ts/radial-gauge/default/files/src/app/__path__/__filePrefix__.test.tsx
```diff
@@ -0,0 +1,8 @@
+import { expect, test } from 'vitest';
+import { render } from '@testing-library/react';
+import $(ClassName) from './index';
+
+test('renders $(ClassName) component', () => {
+    const wrapper = render(<$(ClassName) />);
+    expect(wrapper).toBeTruthy();
+});
```
jackofdiamond5, line 8:
```suggestion
});

```

packages/cli/templates/react/igr-ts/radial-gauge/default/files/src/app/__path__/style.module.css
```diff
@@ -0,0 +1,12 @@
...
+    flex-flow: row;
+    justify-content: space-around;
+}
+
+:local(.title) {
+    color: rgb(0, 153, 255);
+}
+:local(.gauge) {
+    width: 50%;
+}
```
jackofdiamond5, line 12:
```suggestion
}

```

jackofdiamond5 (commented):
Mostly really minor and cosmetic changes and a few notes:

I tried generating a React legacy project, but the cli threw, so we may want to fix that until it is completely removed.

Additionally, the `TypeScriptFileUpdate` for WC & NG and the `ReactTypeScriptFileUpdate` files are in need of some major refactoring (which we may want to prioritize) as they are clones for the most part and we should try to extract some logic into a base file.

jackofdiamond5 (approved):
There's still a little problem with the legacy React projects - because of outdated versions it will fail to compile/run. The problem occurs on newer versions of nodejs (17+) and since we can drop the legacy project entirely (on the next major cli version), there shouldn't be a need to tackle this issue. https://github.com/IgniteUI/igniteui-cli/issues/1176

Lipata:
@ChronosSF, @damyanpetev I'm merging this PR and we should decide if we can release without fixing #1176.

Issue #1165, opened 2023-11-01
Update CLI to use stand-alone components 


ChronosSF:

damyanpetev:
I see in #1174 the current approach is to try to internally split the Ignite UI for Angular project and add an additional prompt for standalone and the `--standalone` flag; While that moves towards a more familiar experience with an option it also requires a bit of extra effort internally since the actual view templates and projects are nothing alike.

Alternative suggestion:
Instead of going for the parity, we could instead plug into the existing setup in the CLI already and define the new standalone setup as a separate project type. That will inheritably plug into the existing prompts here I think:
![image](https://github.com/IgniteUI/igniteui-cli/assets/3198469/5b133dcf-1dda-4f1f-b4c2-a5541a0ed4c4)
And I can see this becoming:
```text
> Ignite UI for Angular
-------------------------
Ignite UI for Angular (non-standalone, legacy)
-------------------------
Ignite UI Angular Wrappers
```
Where the default is obviously standalone, which can be also reflected in the text if we choose to. For calling the `new` command directly outside of Step by step mode, the differentiation will be based on passing a different project `--type` argument.

To that end, in addition I'll suggest we keep the default `igx-ts` type and update/replace it to standalone if possible and assign the existing ones to something like `igx-ts-legacy` (NNF). That means the 
```shell
<cli/schematic> new <name>--framework=angular [--type=igx-ts] --template=side-nav
```
Produces the new Standalone Components setup and is the default option, while:
```shell
<cli/schematic> new <name>--framework=angular --type=igx-ts-legacy --template=side-nav
```
will produce the legacy project with `NgModule`-s.

One last addition - we can and probably should add a migration for existing `igx-ts` project to prompt users if when updating. Something in the line of
```text
> We've updated our view templates to use Standalone Components. Those are compatible with your project and you can start using them now or remain with `NgModule` dependencies. Would you like to continue with Standalone Components for new views? (Y/ n) ? N
We've updated your project type to `igx-ts-legacy` to continue using `NgModule`-s. You can always change back to `igx-ts` should you want to start using Standalone Components.
```
We've checked with @jackofdiamond5 and adding a new standalone view to an existing project works just fine, so not exactly even breaking and we can allow existing projects to use the new templates if they would like to. Minor caveat about versioning can be a limitation with this since schematics follow the library version, depending on how we version it and if users ignore package locks and `ng update` they could jump to the new functionality without notice. But again, not entirely breaking.

[...]

PR #1174, opened 2023-11-17
Update to standalone components

jackofdiamond5:
Closes https://github.com/IgniteUI/igniteui-cli/issues/1165  

Allows the `igniteui-cli` to generate standalone components. This is achieved by separating the templates' structure in two - one for legacy (module-based) apps and one for apps with standalone components. The project structure for standalone components has been reworked so that it matches the project structure of an `ng new` with standalone components.

Every template for every component that's supported by the cli has been updated to be standalone (in the standalone project structure) and has been cleared out of any redundant configurations. 

Routing has been reworked so that now under the `src/app` folder there will be a `app.router.ts` containing all the routes which is then imported in the `app.config.ts` alongside any other global providers. The config file is then provided in the `main.ts` where the application bootstrapping occurs.

Additionally, for all of the custom templates that exist in the cli, the ability to modify a `@Component` metadata had to be introduced since the templates allow the addition of some or all features of the given component, this concerns mainly the custom grids. This caused the need for the cli to be able to add the `imports` property in the `@Component` decorator since now all templates are marked as `standalone` by default and we could not just hardcode all of the required dependencies for these components since this would create too much boilerplate code for any templates that just show a bare bones custom grid for example. So, relying on the template's `dependencies` with their imports marked as `standalone` an additional component transformer was introduced that creates a new `imports` array in the component's meta.

Action list:
- [x] Separate standalone from module-based projects
- [x] Update default standalone projects
    - [x] base
    - [x] base-with-home
    - [x] empty
    - [x] side-nav
    - [x] side-nav-auth
- [x] Update routing for all standalone projects
- [x] Update standalone templates
    - [x] Update custom templates to standalone
- [x] Set up a way to generate standalone & legacy projects through prompt session & the `ig` command
- [ ] Docs


Lipata (approved):

jackofdiamond5:
Currently DockManger throw this error when started:
image

However, this is not a problem with the cli but potentially with the DM package. Since we are initializing the DM in a standalone component, we may need to export the DM component (and potentially some of its internal components) from the package alongside its interfaces, etc. [#ref](https://github.com/IgniteUI/dock-manager/blob/master/src/index.ts)

Lipata (approved):

PR #1189, opened 2024-01-05
Bump the version of WebComponents

valadzhov:
Bumping to 4.7.0 for igniteui-webcomponents and to 4.6.0 for the DV stuff.


Lipata (changes requested):

Lipata:
When I think about it now, it will be valuable if we add templates for the new components.

Lipata (approved):

PR #1198, opened 2024-01-25
feat(cli): upgrade packages for React and WC

### LINKED ISSUE #1191: Implement `ig upgrade` for WebComponents and React
Implement support for `ig upgrade-packages` command in WebComponents and React platforms.
Currently, only Angular is supported.

igdmdimitrov:
Closes #1191


packages/cli/templates/Update.ts
```diff
@@ -0,0 +1,139 @@
...
+	let logicFilesExtension = "";
+	let upgradeable: PackageDefinition[] = [];
+
+	const config = ProjectConfig.getConfig();
+	const framework = config.project.framework;
+	switch (framework.toLowerCase()) {
+		// case "angular":
+		// 	guideLink =
+		// 	  "https://www.infragistics.com/products/ignite-ui-angular/angular/components/general/ignite-ui-licensing";
+		// 	break;
```
gedinakova, line 17: Do we need this commented out chunk?

packages/cli/templates/Update.spec.ts
```diff
@@ -0,0 +1,348 @@
...
+import 'igniteui-react-grids/grids/themes/light/bootstrap.css'
+
+IgrGridModule.register();
+
+export default function Home() {
+	const title = 'igniteui-react example';
+}`,
+				expected:
+`import { something } from 'module';
+import { bait } from 'igniteui-react-core';
```
gedinakova, line 135: Shouldn't 'igniteui-react-core' be licensed?

gedinakova (changes requested):

packages/cli/templates/Update.spec.ts
igdmdimitrov: It is not in the MockFile dependencies and that's why it is not licensed. I'll add it to dependencies in order to be licensed after upgrade.

packages/cli/templates/Update.ts
igdmdimitrov: Depends if we want to keep refactoring and support Angular upgrade here in the same file.

jackofdiamond5 (approved):

gedinakova (approved):

PR #1209, opened 2024-02-20
Provide http client

jackofdiamond5:
Adds the `provideHttpClient` method in the `providers` array for projects with standalone components.

Also sets the project type of legacy angular projects to `igx-ts-legacy` in the cli config.


jackofdiamond5:
Since we've introduced a few hardcoded providers in the base template, we may want to consider adding them only when they are actually needed https://github.com/IgniteUI/igniteui-cli/issues/1210 This should come with some changes in the TSFIleUpdate.

Lipata (approved):

PR #1228, opened 2024-03-20
Generate default redirect routes for NG

jackofdiamond5:
Closes [#30324](https://infragistics.visualstudio.com/Indigo_Platform/_workitems/edit/30324)

Generate default redirect routes for module based & standalone projects.

### DIFF
--- a/packages/core/typescript/TypeScriptFileUpdate.ts
+++ b/packages/core/typescript/TypeScriptFileUpdate.ts
@@ -72,14 +72,22 @@ export class TypeScriptFileUpdate {
 	 * @param linkText Text of the route to add as `data.text`
 	 * @param parentRoutePath Will include the new route as a **child** of the specified route path
 	 * @param routesVariable Name of the array variable holding routes
+	 * @param lazyload Whether to use lazy loading for the route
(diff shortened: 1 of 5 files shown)


packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -78,8 +78,10 @@ export class TypeScriptFileUpdate {
 		routesVariable = DEFAULT_ROUTES_VARIABLE,
 		lazyload = false,
 		routesPath = "",
-		root = false) {
-		this.addRouteModuleEntry(filePath, linkPath, linkText, routesVariable, parentRoutePath, lazyload, routesPath, root);
+		root = false,
```
wnvko, line 81: can we describe in the dev comment above what each parameter is for?

wnvko, line 81: it will be great if parameters are positioned in the same way. Now we have four parameters on one row and all the rest each on its own row :(

wnvko (approved):

wnvko (approved):

Lipata (approved):

PR #1234, opened 2024-04-04
Remove igniteui-react related dependencies

### LINKED ISSUE #1233: Remove igniteui-react related dependencies

igdmdimitrov:
Closes #1233


packages/cli/templates/react/igr-ts/projects/_base/files/package.json
```diff
@@ -8,9 +8,6 @@
     "@testing-library/react": "^14.0.0",
     "element-internals-polyfill": "^1.3.10",
     "functions-have-names": "^1.2.3",
-    "igniteui-dockmanager": "^1.13.0",
-    "igniteui-react": "18.3.0",
-    "igniteui-react-core": "18.3.0",
```
wnvko, line 13: if we are removing the core here, we should add it then to all the project which are using it, e.g. pie-chart, grid basic, category chart and so on.
_Note: after creating a react project with the CLI, added some components, and then tried to install manually igniteui-react-core@18.6.0 it failed with `Cannot read properties of undefined (reading 'fileExists')`!_

wnvko (changes requested):
Add core package to all templates that need it.

packages/cli/templates/react/igr-ts/projects/_base/files/package.json
damyanpetev: That error is likely from the CLI packages (check the full stack), not really related to the react-core package.
The rest is true though, if that package is a dependency of some (or all) components, views using them should have it listed as a dependency to install.

wnvko (approved):

packages/cli/templates/react/igr-ts/radial-gauge/default/index.ts
```diff
@@ -12,7 +12,7 @@ class IgrTsRadialGaugeTemplate extends IgniteUIForReactTemplate {
 		this.description = `provides a number of visual elements, like a needle, tick marks, ranges
 							and labels, in order to create a predefined shape and scale.`;
 		// TODO: read version from igniteui-react-core in package.json
-		this.packages = ["igniteui-react-gauges@18.3.0"];
+		this.packages = ["igniteui-react-gauges@~18.6.0"];
```
damyanpetev, line 15: @igdmdimitrov  This can't be right if `igniteui-react-core` is not auto-installed, since `igniteui-react-gauges` 
**requires** it still: https://cdn.jsdelivr.net/npm/igniteui-react-gauges/package.json
Same for charts, grids and pretty much all the react packages I think except `igniteui-react`

damyanpetev: Nvm, it seems peer deps are also auto-installed actually, so not really needed to be explicitly listed as dep if not useful for the rest of the app, i.e. the core.

PR #1235, opened 2024-04-04
chore: update typescript; migrate to eslint

jackofdiamond5:
Closes [#30607](https://infragistics.visualstudio.com/Indigo_Platform/_workitems/edit/30607)

This PR also migrates the linter that we use to `eslint` and applies the appropriate configuration. This caused the need to extract the `eslintconfig` for some of the templates (namely React) outside of the `package.json` as that caused `eslint` to attempt and extend the provided configurations which (as expected) were not installed.

Some additional changes were needed in the TSFileUpdate/Utils to reflect some updates to TS API.


wnvko (approved):

packages/core/prompt/BasePromptSession.ts
```diff
@@ -408,6 +408,10 @@ export abstract class BasePromptSession {
 				}
 			}
 
+			if (config.project.framework === "react") {
```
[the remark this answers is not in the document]
jackofdiamond5: This is how we used to do it, yes. However, now that we are using `eslint` having an `extend` option in the `package.json` causes it to attempt to extend the provided modules. These modules are part of the templated projects and even though I've specified that we strictly want to ignore directories under the templates, `eslint` still goes over them and looks at their configs. And since the package is nowhere to be found (as you'd expected from the templates' dependencies) `eslint` throws. So, I managed to work around this by adding this option dynamically when the project structure is present. I've described this briefly in the PR's description but maybe I should've been more thorough. :smile:

packages/core/util/Util.ts
```diff
@@ -545,6 +553,27 @@ export class Util {
...
+				packageJson.eslintConfig.extends = ["react-app"];
+				if (typeof configOrPath === "object"
+					&& configOrPath.project.projectType === "es6") {
+					packageJson.eslintConfig.extends.push("react-app/jest");
+				}
+				this.writeFile(pkgJson, JSON.stringify(packageJson, null, 2));
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
jackofdiamond5: https://github.com/IgniteUI/igniteui-cli/pull/1235#issue-2225742206, https://github.com/IgniteUI/igniteui-cli/pull/1235#discussion_r1553488006

.eslintrc.json
```diff
@@ -0,0 +1,155 @@
...
+	"root": true,
+	"env": {
+		"browser": true,
+		"es6": true,
+		"node": true
+	},
+	"extends": [
+		// TODO: consider extending the recommended by ts rules and overriding them where necessary
+		// 'eslint:recommended',
+		// 'plugin:@typescript-eslint/recommended'
```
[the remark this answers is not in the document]
jackofdiamond5: After this PR is merged, I will log an issue to dig deeper into eslint as this was not the idea of the PR, it just happened because tslint was throwing after I upgraded ts and the packages around it.

packages/core/util/Util.ts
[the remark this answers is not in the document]
damyanpetev: .eslintrc.json -> `_dot_eslintrc.json` separate file

packages/cli/lib/commands/new.ts
```diff
@@ -1,4 +1,4 @@
-import { GoogleAnalytics, PackageManager, ProjectConfig, ProjectLibrary, Util } from "@igniteui/cli-core";
+import { Config, GoogleAnalytics, PackageManager, ProjectConfig, ProjectLibrary, Util } from "@igniteui/cli-core";
```
damyanpetev, line 1: Leftover at this point I assume

packages/cli/package.json
```diff
@@ -78,16 +78,16 @@
     "all": true
   },
   "dependencies": {
-    "@igniteui/angular-templates": "~17.1.13112",
-    "@igniteui/cli-core": "~13.1.12",
+    "@igniteui/angular-templates": "~17.1.13113-beta.6",
+    "@igniteui/cli-core": "~13.1.13-beta.6",
```
damyanpetev, line 82: Are those supposed to be changed by this PR? We usually change versions is a singular version commit before tagging, no?

Lipata: It's needed when we tag beta versions from a feat branch/PR not from master

packages/cli/templates/react/igr-es6/projects/_base/index.ts
```diff
@@ -1,4 +1,5 @@
-import { ControlExtraConfiguration, defaultDelimiters, ProjectTemplate, Util } from "@igniteui/cli-core";
+import { App, ControlExtraConfiguration, defaultDelimiters,
+	FS_TOKEN, IFileSystem, ProjectTemplate, Util } from "@igniteui/cli-core";
```
damyanpetev, line 2: Same here

packages/cli/package.json
damyanpetev: Still, don't care for that to be merged in master. Could consider changing the release process if you'll keep tagging from branches and have the release action set the version from the tag or something

packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -36,6 +36,12 @@ export class TypeScriptFileUpdate {
 		this.initState();
 	}
 
+	public transform(source: ts.SourceFile | ts.SourceFile[],
```
damyanpetev, line 39: What's the point of this when all it's doing is acting as a proxy of the same method?
And also causes the usage code to be forced to type assert as well, since this isn't typed quite like the original TS method. IF we have to keep this, do also copy the generic typing from the wrapped method and drop all the extra type asserts you were forced to add

packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -211,7 +217,7 @@ export class TypeScriptFileUpdate {
 						...newProvides
 					]);
 
-					return ts.factory.updateArrayLiteralExpression(array, elements);
+					return ts.factory.updateArrayLiteralExpression(array, elements.map(x => x as ts.Expression));
```
damyanpetev, line 220: Not C#, looping over a collection to change the items type is a redundant operation, so please don't :)
You _can_ simply type assert `elements as ts.Expression[]` OR even better - how about just above when this is created we specify the item type there:
```ts
ts.factory.createNodeArray<ts.Expression>(...
```

packages/core/util/FileSystem.ts
```diff
@@ -12,9 +12,9 @@ export class FsFileSystem implements IFileSystem {
 			return false;
 		}
 	}
-	public readFile(filePath: string, encoding?: string): string {
+	public readFile(filePath: string, encoding?: BufferEncoding): string {
```
damyanpetev, line 15: Uf, this should... also reflect in the interface? Since it's quite useless in the actual implementation that nothing should refer directly.

Should we consider just defaulting to UTF8 in general?

packages/core/util/Util.ts
```diff
@@ -43,6 +43,14 @@ export class Util {
 		return App.container.get<IFileSystem>(FS_TOKEN).fileExists(filePath);
 	}
 
+	public static readFile(filePath) {
```
damyanpetev, line 46: 👀 Okay, this is a bit confusing. Why the two specific shortcuts, when the setup is intended for all usage to just grab the fs from the container and use it as needed?

spec/acceptance/add-spec.ts
```diff
@@ -286,7 +286,7 @@ describe("Add command", () => {
 			expect(fs.existsSync(componentPath)).toBeTruthy();
 			// file contents:
 			expect(fs.readFileSync(componentPath, "utf-8")).toContain("export class TestViewComponent");
-			expect(fs.readFileSync("src/app/app-routing.module.ts", "utf-8")).toBe(
+			expect(fs.readFileSync("src/app/app-routing.module.ts", "utf-8").replace(/(\r?\n)+/g, EOL)).toBe(
```
damyanpetev, line 289: Hm these tests were quite specific, what happened here? TS formatting service changed defaults?

damyanpetev (changes requested):
Some questions, this is mostly for the typescript changes

packages/core/typescript/TypeScriptFileUpdate.ts
jackofdiamond5: This is a proxy that is currently used for the tests. It will be gone with https://github.com/IgniteUI/igniteui-cli/pull/1238

packages/core/util/Util.ts
```diff
@@ -5,7 +5,7 @@ import * as glob from "glob";
 import * as path from "path";
 import through2 = require("through2");
 import { BaseComponent } from "../templates/BaseComponent";
-import { Component, ComponentGroup, Delimiter, FS_TOKEN, IFileSystem, Template, TemplateDelimiters } from "../types";
+import { Component, ComponentGroup, Config, Delimiter, FS_TOKEN, IFileSystem, Template, TemplateDelimiters } from "../types";
```
damyanpetev, line 8:
```suggestion
import { Component, ComponentGroup, Delimiter, FS_TOKEN, IFileSystem, Template, TemplateDelimiters } from "../types";
```

packages/ng-schematics/package.json
```diff
@@ -20,16 +20,16 @@
   "dependencies": {
     "@angular-devkit/core": "~14.0.0",
     "@angular-devkit/schematics": "~14.0.0",
```
damyanpetev, line 22: @Lipata you should really update those too

spec/acceptance/add-spec.ts
```diff
@@ -363,13 +363,13 @@ export class AppModule {
 			// file contents:
 			expect(fs.readFileSync(componentPath, "utf-8")).toContain("export class TestNestedFoldersComponent");
 
-			expect(fs.readFileSync("src/app/app-routing.module.ts", "utf-8")).toBe(
+			expect(fs.readFileSync("src/app/app-routing.module.ts", "utf-8").replace(/(\r?\n)+/g, EOL)).toBe(
```
damyanpetev, line 366:
```suggestion
			expect(fs.readFileSync("src/app/app-routing.module.ts", "utf-8")).toBe(
```

spec/acceptance/add-spec.ts
```diff
@@ -363,13 +363,13 @@ export class AppModule {
...
 				`import { TestNestedFoldersComponent } from './${componentFolder}/test-nested-folders.component';` + EOL +
 				// tslint:disable-next-line:max-line-length
 				`const routes: Routes = [{ path: 'test-nested-folders', component: TestNestedFoldersComponent, data: { text: 'Test Nested Folders' } }];` + EOL
 			);
 
-			expect(fs.readFileSync("src/app/app.module.ts", "utf-8")).toBe(
+			expect(fs.readFileSync("src/app/app.module.ts", "utf-8").replace(/(\r?\n)+/g, EOL)).toBe(
```
damyanpetev, line 372:
```suggestion
			expect(fs.readFileSync("src/app/app.module.ts", "utf-8")).toBe(
```

damyanpetev (approved):

Lipata (approved):

PR #1237, opened 2024-04-08
fix(cli-templetes): Replaced index.html <title> to ApplicationTitle as placeholder

ivanvpetrov:
Replaced React's, Web Component's and Angular's template index.html <title> to ApplicationTitle as placeholder


packages/cli/templates/react/igr-ts/projects/_base/files/index.html
```diff
@@ -8,7 +8,7 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
     <meta name="theme-color" content="#000000" />
-    <title>IgniteUI for React</title>
+    <title><%=ApplicationTitle%></title>
```
onlyexeption, line 11: this causes the placeholder to appear in other templates (empty and top-nav) because they extend base with home that doesn't have its own index.html which means that it will reuse the one from _base containing the placeholder
image

packages/igx-templates/igx-ts-legacy/projects/_base/index.ts
```diff
@@ -65,7 +65,8 @@ $app-palette: palette($primary, $secondary, $surface);
 			dot: ".",
 			path: name,
 			projectTemplate: this.id,
-			yamlDefaultBranch: "<%=yaml-default-branch%>" // the placeholder will be evaluated by CodeGen
+			yamlDefaultBranch: "<%=yaml-default-branch%>", // the placeholder will be evaluated by CodeGen
+			ApplicationTitle: "<%=ApplicationTitle%>" // the placeholder will be evaluated by CodeGen
```
onlyexeption, line 69: since this is used only in base template, I dont see a reason to have this in the config. It basically evaluates the <%=ApplicationTitle%> placeholder as <%=ApplicationTitle%>, also please use camelCase. This applies for igx-ts as well

onlyexeption:
please use camelCase across your changes

packages/cli/templates/react/igr-ts/projects/_base/files/index.html
ivanvpetrov: A new index.html with 'IgniteUI for React' title is added in packages\cli\templates\react\igr-ts\projects\_base_with_home\files. 
This way it overrides the one from base and won't leak the <%=ApplicationTitle%> placeholder to other templates anymore.

packages/igx-templates/igx-ts-legacy/projects/_base/index.ts
ivanvpetrov: As we discussed, currently there is no concreate naming convention for placeholders for the project. There are instances of PascalCase, camelCase, kebab-case and snake_case. 

And since in the mentioned code, where we put  [ApplicationTitle: "<%=ApplicationTitle%>"], there are already props with capital letters (e.g CustomTheme, DefaultTheme), we've decided to leave it like this and not change it, because this will force changes in several other places with no benefit.

onlyexeption (approved):

gedinakova (approved):

PR #1238, opened 2024-04-08
Implement an AST transformer

jackofdiamond5:
Closes [#30400](https://infragistics.visualstudio.com/Indigo_Platform/_workitems/edit/30400).  

Docs - [TypeScriptASTTransformer Wiki](https://github.com/IgniteUI/igniteui-cli/wiki/TypeScriptASTTransformer)


packages/core/typescript/TypeScriptFormattingService.ts
```diff
@@ -0,0 +1,217 @@
...
+  /**
+   * The printer instance to use to get the source code from the AST.
+   */
+  private get printer(): ts.Printer {
+    if (!this._printer) {
+      this._printer = ts.createPrinter(this.printerOptions);
+    }
+
+    return this._printer;
+  }
```
damyanpetev, line 33: That's weird the service should need that

packages/core/typescript/TypeScriptFormattingService.ts
```diff
@@ -0,0 +1,217 @@
...
+    if (this.formatOptions.singleQuotes) {
+      this.sourceFile = ts.transform(
+        this.sourceFile,
+        [this.convertQuotesTransformer],
+        this.compilerOptions
+      ).transformed[0] as ts.SourceFile;
+    }
+
+    const text = this.applyChanges(
+      this.printer.printFile(this.sourceFile),
```
damyanpetev, line 70: Huh? There really is no point re-printing the file over and over just so this service can work with `ts.SourceFile`.
How about we give it the path it needs anyway and the formatting just reads and manipulates the text as intended?

packages/core/typescript/TypeScriptFormattingService.ts
```diff
@@ -0,0 +1,217 @@
...
+    (rootNode: T): ts.Node => {
+      const visit = (node: ts.Node): ts.Node => {
+        if (ts.isStringLiteral(node)) {
+          const text = node.text;
+          // the ts.StringLiteral node has a `singleQuote` property that's not part of the public APi for some reason
+          // to make our lives easier we can modify it though
+          const singleQuote = (node as any).singleQuote;
+          const newNode = context.factory.createStringLiteral(text);
+          (newNode as any).singleQuote =
+            singleQuote || this.formatOptions.singleQuotes;
```
damyanpetev, line 155: Nope. The fact this is not exposed should say something.

Also, notice how the old logic only formats `createdStringLiterals`, i.e. the literals _we_ added to the source, the rest is _not_ the responsibility of the transform.
Also also, funnily enough, that was likely before `createStringLiteral` had a second param to specify that? Because we do have it used in places and it does work:
image
---
So, how about we create the literals right the first time and skip the entire formatting quotes bit. Should also allow you to drop the transform step and the printer

packages/core/typescript/TypeScriptFormattingService.ts
damyanpetev, line 1: Also since this is more/less equivalent of the old formatting section, would love to see it replaced even in this PR. Bit hard to track what functionality is being moved otherwise.

damyanpetev (commented):
Formatting service pass

packages/core/typescript/TypeScriptFormattingService.ts
```diff
@@ -0,0 +1,217 @@
...
+   * Create a new formatting service for the given source file.
+   * @param sourceFile The source file to format.
+   * @param fileSystem The file system to use when reading formatting options.
+   * @param formatSettings Custom formatting settings to apply.
+   * @param printerOptions Options to use when printing the source file.
+   * @param compilerOptions Compiler options to use when transforming the source file.
+   */
+  constructor(
+    public sourceFile: ts.SourceFile,
+    private readonly fileSystem?: IFileSystem,
```
damyanpetev, line 45: Also, that's kind of the point of the App container - you don't have to pass the FS around

packages/core/typescript/TypeScriptASTTransformer.ts
```diff
@@ -0,0 +1,771 @@
...
+   * @remarks The `anchorElement` must match the type of the elements in the collection.
+   */
+  public addMembersToArrayLiteral(
+    visitCondition: (node: ts.Node) => boolean,
+    propertyAssignment: IPropertyAssignment[],
+    prepend?: boolean,
+    anchorElement?: ts.Expression | IPropertyAssignment
+  ): ts.SourceFile;
+  public addMembersToArrayLiteral(
+    visitCondition: (node: ts.Node) => boolean,
```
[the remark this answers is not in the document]
damyanpetev: However, I'm afraid that as neat as those methods are, they just won't work for us in this format.
The example in the description as given:
```ts
this.addMembersToArrayLiteral(node =>
    ts.isArrayLiteralExpression(node) &&
    node.elements.some(e =>
        ts.isObjectLiteralExpression(e) &&
        e.properties.some(
          (p) => ts.isPropertyAssignment(p) && p.name.getText() === "path"
        )
    ),
  // the elements to be added to the array literal
  [newObjectLiteral]
);
```
breaks down the moment it's applied to this:
```ts
const filesConfigs: [
  { file: "someConfig", path: "someUrl";
];

@mappingDecorator({ aliases: [{ path: "/endpoint"}, { path: "/oldendpoint" ] })
public endPoint() {}

const routes = [
   { path: "/", component: Home }
];
```
Would it add to all or just the first? You get the point, right? There's a reason we look for specific decorators by name with child properties or explicit variable names like `routes` and so on.

PS: Somewhat fixed formatting, please whatever your prettier settings are... fix em, presumably the max line at 80 just causes silly formatting all over

packages/core/typescript/TypeScriptFormattingService.ts
```diff
@@ -0,0 +1,217 @@
...
+import { IFormattingService, IFormatSettings } from '../types';
+import { IFileSystem } from '../types/FileSystem';
+import { Util } from '../util';
+
+const NEW_LINE_PLACEHOLDER = '//I keep the new line';
+export class FormattingService implements IFormattingService {
+  private _printer: ts.Printer | undefined;
+  private _formatSettingsFromConfig: IFormatSettings = {};
+  private _defaultFormatSettings: IFormatSettings = {
+    indentSize: 3,
```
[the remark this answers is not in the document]
jackofdiamond5: Ah, this one should be a `4`.

packages/core/typescript/TypeScriptFormattingService.ts
```diff
@@ -0,0 +1,217 @@
+import * as ts from 'typescript';
+import { IFormattingService, IFormatSettings } from '../types';
+import { IFileSystem } from '../types/FileSystem';
+import { Util } from '../util';
+
+const NEW_LINE_PLACEHOLDER = '//I keep the new line';
```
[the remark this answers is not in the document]
jackofdiamond5: Can't quite agree there. The transformer is not a formatter, and it shouldn't be concerned with applying any sorts of format options to the result after parsing the AST. I want to make these two completely separate entities. This could mean that _a)_ the transformer does not print anything on finalize unless a formatter is provided or _b)_ the transformer does print something, but it's not formatted. I chose the latter to keep the finalize method's signature consistent and usable for unit test where we want to just test if the transformer has done its thing without having to explicitly provide a formatter as well.

Regarding the utils I agree that the `getFileSource` and `saveFile` functions are rather useful and do exactly what we want, and I'll be using them in the platform-specific transformers/file updates/whatever you call them and here in the formatter. Though, I should mention that this refactoring effort is not to stop with the file updates refactoring, there's a bigger fish to fry later on when we start implementing our own `vfs`. I have some specific plans for that (which we need to talk about more) and I've been doing this refactoring while keeping in mind what's coming next. Though I agree that some of the decisions made here (like making the formatter use `ts.SourceFIle`) don't make much sense if you don't consider our own `vfs`. So, for the moment I can make it use the regular `App.container` and change that later on if that's fine with you.

packages/core/typescript/TypeScriptFormattingService.ts
jackofdiamond5: This is a good idea, but it would require that the transformer knows about the formatting options and has access to the file system (to get `.editorconfig/eslint` options) which is precisely what I want to avoid as an AST transformer should be concerned with the AST only and not the file system. I could do what we used to be doing - keep track of all the literals that we create and then replace them in the formatter based on the format options or I could try to figure out a different way to implement the transformer.

packages/core/typescript/TypeScriptFormattingService.ts
jackofdiamond5: I agree that with the current implementation it makes more sense to pass in the `path/to/file` that will use the `App.container` to read and apply formatting.

packages/core/typescript/TypeScriptFormattingService.ts
jackofdiamond5: If the service does not operate on a `ts.SourceFIle` but reads the content from the FS, then yes it does not really need a printer.

Lipata (approved):

PR #1252, opened 2024-05-27
chore(ng): update to ng 18

Lipata:
Closes https://github.com/IgniteUI/igniteui-cli/issues/1253


packages/ng-schematics/src/utils/NgFileSystem.ts
```diff
@@ -24,13 +24,18 @@ export class NgTreeFileSystem implements IFileSystem {
...
+	 * @param ignorePattern Optional pattern to ignore
 	 */
-	public glob(dirPath: string, pattern: string): string[] {
+	public glob(dirPath: string, pattern: string, ignorePattern?: string): string[] {
 		const dir = this.tree.getDir(dirPath);
 		const entries: string[] = [];
 		pattern = pattern.split("**/*").pop() || pattern;
 
 		dir.visit((_fullPath, entry) => {
+			if (ignorePattern && entry?.path.includes(ignorePattern)) {
```
jackofdiamond5, line 35: This here is a rather naive fix for https://github.com/IgniteUI/igniteui-cli/issues/1253. The underlying issue is that `glob` is called for the root of a project and it traverses the `node_modules` in an attempt to find `package.json` files. It succeeds a bit too well in that regard, and the following logic throws due to various assumptions made for the specific root-level `package.json` file. While this change should fix the issue, it is a slow solution as the `DirEntry` visitor still loops through the `node_modules`. I'm thinking, we should modify the `glob` function to build dir-entries as roots for each subfolder that does not match the ignore pattern and then invoke the visitor on each of them separately, populating the entries array in the process, just at thought at this point.

PR #1254, opened 2024-05-31
chore: release 13.3.0

Lipata:


jackofdiamond5 (approved):

packages/cli/lib/commands/upgrade.ts
```diff
@@ -55,7 +55,7 @@ command = {
 					if (success && !argv.skipInstall) {
 						Util.log("Installing npm packages.");
 						try {
-							Util.execSync("npm install --quiet");
+							Util.execSync("npm install --force --quiet");
```
jackofdiamond5, line 58: Maybe `--legacy-peer-deps` instead of `--force` but alright for the time being.

PR #1286, opened 2024-07-23
add test files for app component in WC and Angular 

onlyexeption:
Closes [30492](https://infragistics.visualstudio.com/Indigo_Platform/_workitems/edit/30492/) in CodeGen


packages/igx-templates/igx-ts-legacy/projects/_base/files/src/app/app.component.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { AppComponent } from './app.component';
+
+describe('MasterViewComponent', () => {
```
wnvko, line 4: `AppComponent` instead of `MasterViewComponent`?

packages/igx-templates/igx-ts/projects/_base/files/src/app/app.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
+import { AppComponent } from './app.component';
+
+describe('MasterViewComponent', () => {
```
wnvko, line 4: `AppComponent` instead of `MasterViewComponent`?

wnvko (changes requested):
Adding spec file to `_base` project will add this file to all the projects in `projects` folder both for Angular and WebComponents. All these projects imports routing and some other ignite-ui components which are not added in the spec file. IMO for each project we should create specific spec file.

onlyexeption:
> Adding spec file to `_base` project will add this file to all the projects in `projects` folder both for Angular and WebComponents. All these projects imports routing and some other ignite-ui components which are not added in the spec file. IMO for each project we should create specific spec file.

For igx-ts and igx-ts-legacy:

I added RouterTestingModule to the test imports in the base template. Now the `empty` and `base` (_base and _base_with_home) projects will have it imported in their spec files.  The side-nav and side-nav-auth have their own app spec files.

For webcomponents, I double checked different component's test files as well as the CodeGen template. They have the same structure everywhere so I don't think its necessary to create separate app.test.ts files.

jackofdiamond5 (approved):

Lipata (approved):

PR #1297, opened 2024-08-01
Providing class name to routing if we don't want to use the first class in file

dafo:


packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -90,13 +90,14 @@ export class TypeScriptFileUpdate {
 		linkPath: string,
 		linkText: string,
 		parentRoutePath: string,
+		className: string,
```
jackofdiamond5, line 93: Add this as an optional argument at the very end of the arguments list.

packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -313,7 +316,11 @@ export class TypeScriptFileUpdate {
 		let className: string;
 		const fileSource = TsUtils.getFileSource(filePath);
 		const relativePath: string = Util.relativePath(this.targetPath, filePath, true, true);
-		className = TsUtils.getClassName(fileSource.getChildren());
+		if (providedClassName != "") {
+            className = providedClassName;
+        } else {
+            className = TsUtils.getClassName(fileSource.getChildren());
```
jackofdiamond5, line 322: Probably a better idea is to do `className = providedClassName ||  TsUtils.getClassName(fileSource.getChildren());` and ditch the rest of the `if` statement. Since that would make sure any `falsy` value is accounted for. With the current approach, if I pass in values such as `undefined` or `null`, that would cause the runtime to enter in the first block of the `if` statement since both are technically not empty strings. Which would result in an empty `className` and potential errors afterwards.

packages/core/typescript/TypeScriptFileUpdate.ts
```diff
@@ -308,12 +309,12 @@ export class TypeScriptFileUpdate {
...
 		root = false,
-		isDefault = false
+		isDefault = false,
+		providedClassName?: string
 	) {
-		let className: string;
 		const fileSource = TsUtils.getFileSource(filePath);
 		const relativePath: string = Util.relativePath(this.targetPath, filePath, true, true);
-		className = TsUtils.getClassName(fileSource.getChildren());
+		const className = providedClassName ||  TsUtils.getClassName(fileSource.getChildren())
```
jackofdiamond5, line 317:
```suggestion
		const className = providedClassName || TsUtils.getClassName(fileSource.getChildren());
```

jackofdiamond5 (approved):

Lipata (approved):

Issue #1294, opened 2024-08-01
Update the CLI's dependencies


jackofdiamond5:
### Description  

We need to update the CLI's packages as they are heavily outdated. This task is bound to introduce refactoring in the code as the updated packages can/will have different API.  Updating the `inquirer` and `jasmine` could be rather time consuming due to the amount of code that will have to be changed.

Part two of this task is to introduce strict typing everywhere possible in the CLI and minimize the use of `any` as much as possible. Also, we should revisit some types that are not really used/have duplicates, and clean them up.

This task is a prerequisite for:
 - #1305
 - #1236.

### Checklist:

- [x] Root project & CLI package
    - [x] Utils (chalk, glob, lite-server, etc.)
    - [x] yargs - with updates to all CLI commands 
    - [x] inquirer
- [x] Templates - no changes were needed
- [x] Schematics (jasmine, node, devkit)
- [x]  jasmine for `ng-schematics`, `cli` & root project
- [x] Core (through2, devkit)
- [x] Typescript - upgrade to 5.5.4

PR #1311, opened 2024-09-04
Dependency updates iteration 1

jackofdiamond5:
Related to https://github.com/IgniteUI/igniteui-cli/issues/1294

This PR updates packages of the root CLI project, as well as the CLI package sub-project.

Updating `yargs` required some changes in all of the CLI's commands, as the newest version of the package is used and the API is quite different to what we've been using so far. Additionally, strict types have been introduced everywhere possible for each individual command. Also, some new tests have been introduced for the `help` command since we did not have full coverage for all available commands. Finally, the `fs-extra` package has been removed and the code has been updated to use `fs` instead,

Currently there is a problem with the CLI where running a command with aliased options fails - https://github.com/IgniteUI/igniteui-cli/issues/1312

Notes for testing:
- All commands should be tested one at a time to ensure that they work as expected.
  - Checklist:
     - [x] add command
     - [x] build command
     - [x] config command
     - [x] doc command
     - [x] generate command
     - [x] list command
     - [x] help command
         - [x] global (e.g. `ig --help`)
         - [x] for individual commands (e.g. `ig add --help`)
     - [x] new command
     - [x] quickstart command
     - [x] start command
     - [x] test command
     - [x] upgrade command

- A project should be generated with `sass`/`less` styles and it should work just fine (all styles files should be copied properly) - this is due to the removal of `fs-extra` where it was used to copy the styles to the generated project's directory. Now custom logic handles this and it should be tested.


Hristo313 (approved):

Lipata (approved):

packages/cli/lib/commands/config.ts
```diff
@@ -1,60 +1,68 @@
 import { Config, GoogleAnalytics, ProjectConfig, Util } from "@igniteui/cli-core";
+import { ConfigCommandType, PositionalArgs } from "./types";
+import { ArgumentsCamelCase } from "yargs";
```
dafo, line 3:
```suggestion
```

packages/cli/lib/commands/config.ts
```diff
@@ -1,60 +1,68 @@
...
+					.option("value", {
+						describe: "New value to add",
+						type: "string"
+					})
+					.usage(""); // do not show any usage instructions before the commands list
 				},
-				value: {
-					describe: "New value to add",
-					type: "string"
+				handler: (argv: ArgumentsCamelCase<PositionalArgs>) => {
```
dafo, line 52:
```suggestion
				handler: command.addHandler
```

packages/cli/lib/commands/config.ts
```diff
@@ -1,60 +1,68 @@
...
+					command.addHandler(argv);
```
dafo, line 53:
```suggestion
```

Lipata (approved):

Hristo313 (approved):

PR #1315, opened 2024-09-05
chore(react): increase build ram

Lipata:


jackofdiamond5 (approved):
This is a very temporary fix and following https://github.com/vitejs/vite/issues/2433, we may want to try and minimize whatever we are building, though in the case of the charts that might be difficult. We could try to split the building of the DV components from the rest of the app and also avoid generating source maps as they tend take a lot of memory.

PR #1321, opened 2024-10-03
chore: update wc, react and dock-manager

Lipata:
Update old references in the templates


valadzhov (approved):

packages/cli/templates/webcomponents/igc-ts/projects/_base/files/package.json
```diff
@@ -21,9 +21,9 @@
   },
   "dependencies": {
     "@vaadin/router": "^1.7.4",
-    "lit": "^3.1.2",
-    "typescript": "^5.3.3",
-    "igniteui-webcomponents": "~4.11.0"
+    "lit": "^3.2.0",
```
jackofdiamond5, line 24: For WC I get 
image

This is because the `igc-form` component has been removed in https://github.com/IgniteUI/igniteui-webcomponents/pull/1302 so we need to use the native form in our templates.

jackofdiamond5 (changes requested):

jackofdiamond5 (changes requested):
For WC I get 
image

This is because the `igc-form` component has been removed in https://github.com/IgniteUI/igniteui-webcomponents/pull/1281, https://github.com/IgniteUI/igniteui-webcomponents/pull/1302 so we need to use the native form in our templates.

jackofdiamond5 (approved):
I found a little problem with the radio group, where you can select more than one option. But this is not necessarily related to this PR.

PR #1322, opened 2024-10-03
chore: update wc, react and dock-manager

Lipata:
Update old references in the templates


jackofdiamond5 (changes requested):
The comment on this PR https://github.com/IgniteUI/igniteui-cli/pull/1321#discussion_r1787775650 applies here as well.

jackofdiamond5 (approved):

PR #1323, opened 2024-10-11
Allow reordering of array literal expressions

jackofdiamond5:
Closes [AzureDevOps#34206](https://infragistics.visualstudio.com/Indigo_Platform/_workitems/edit/34206)


wnvko (dismissed):
In the end we will sort all the IGX things after everything else. Hope this will not break anything.

valadzhov:
The issue is not reproducible with this ordering but it is not the desired one where the Igx items are last.
image

valadzhov:
This logic does not sort the imports of projects created with the CLI, it is related to the sorting of imports of projects created with the CodeGen and since this was tested yesterday I changed the status of the task to "Completed".

PR #1329, opened 2024-10-28
fix(ng): use standalone tsconfig for legacy modules projects

Lipata:
If possible to use the same tsconfig


Lipata:
`tsconfigs` will be whatever `angular-cli` is generated for `standalone` & `ngModules` respectively.

PR #1331, opened 2024-11-04
resolve code scanning vulnerability alerts

Hristo313:
Resolve Vulnerability alerts in the igniteui-cli

Code scanning alerts fixes:
- Dismiss the insecure randomness alerts because this is only used for dummy data, and there is no need for security in this case.
- Replace execSync with spawnSync to avoid constructing the shell command by concatenating strings with untrusted input.
- Add check to ensure that the function does not allow merging of special properties like proto and constructor.

### DIFF
--- a/packages/cli/templates/webcomponents/igc-ts/grid/grid-editing/files/src/app/__path__/DataGridSharedData.ts
+++ b/packages/cli/templates/webcomponents/igc-ts/grid/grid-editing/files/src/app/__path__/DataGridSharedData.ts
@@ -9,7 +9,7 @@ export class DataGridSharedData {
     let maleCount: number = 0;
     let femaleCount: number = 0;
     for (let i = 0; i < count; i += 1) {
-      const age: number = Math.round(this.getRandomNumber(20, 40));
+      const age: number = Math.round(this.getSecureRandomNumber(20, 40));
       const gender: string = this.getRandomGender();
       const firstName: string = this.getRandomNameFirst(gender);
       const lastName: string = this.getRandomNameLast();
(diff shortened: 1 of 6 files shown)


packages/cli/templates/webcomponents/igc-ts/grid/grid-editing/files/src/app/__path__/DataGridSharedData.ts
```diff
@@ -9,7 +9,7 @@ export class DataGridSharedData {
     let maleCount: number = 0;
     let femaleCount: number = 0;
     for (let i = 0; i < count; i += 1) {
-      const age: number = Math.round(this.getRandomNumber(20, 40));
+      const age: number = Math.round(this.getSecureRandomNumber(20, 40));
```
kdinev, line 12: I'm pretty sure that these about the random numbers in dummy data can safely be marked as false positives.

Lipata (approved):

packages/core/packages/PackageManager.ts
```diff
@@ -123,18 +123,18 @@ export class PackageManager {
...
 			/* passes through */
 			default:
-				command = `${managerCommand} uninstall ${packageName} --quiet --save`;
+				args = ['uninstall', packageName, '--quiet', '--save'];
 				break;
 		}
 		try {
 			// tslint:disable-next-line:object-literal-sort-keys
-			Util.execSync(command, { stdio: "pipe", encoding: "utf8" });
+			spawnSync(managerCommand, args, { stdio: "pipe", encoding: "utf8" });
```
Lipata, line 137: I wonder if its worth adding spawnSync to Util, and wrap it in a try-catch, like it is with execSync?

Lipata (changes requested):

Lipata (approved):

PR #1334, opened 2024-11-06
remove lite-server package and quickstart

Hristo313:
Resolve Vulnerability alerts in the igniteui-cli

Dependabot alerts fixes:
- remove lite-server package
- remove quickstart


package.json
```diff
@@ -93,7 +91,9 @@
 		"micromatch": "^4.0.8",
 		"trim-newlines": "^3.0.1",
 		"parse-path": "^5.0.0",
-		"http-cache-semantics": "^4.1.1"
+		"http-cache-semantics": "^4.1.1",
+		"send": "^0.19.0",
```
jackofdiamond5, line 95: We may not be needing `send` or `axios` as dependencies since I can't find any logic in the CLI that uses them.

Hristo313: Done

jackofdiamond5 (approved):

PR #1364, opened 2025-01-13
Replace rollup with vite configuration

### LINKED ISSUE #1052: Fix the WC project not getting tree-shaking with rollup (or move to webpack / other module bundler)
The WC project doesn't receive tree-shaking. It's easy to reproduce if you debug a WC project made with the CLI and see that all the js files are included in the bundle by default.

If you add the separate views in a create-react-app project instead, treeshaking is achieved by webpack so it is definitely possible.

This is either some issue in rollup's config or maybe more generic inability for rollup to treeshake the WC project. In this case we might need to use another bundler.

Hristo313:
Closes #1052

Remove rollup for web components projects and add vite configuration.
Update versions of typescript, concurrently, rimraf, tslib, lit.


Lipata (changes requested):

rkaraivanov:
@Hristo313 @Lipata 

IMO you assuming this is going to be a Lit based application I highly recommend mimicking what the default Vite [lit-ts](https://github.com/vitejs/vite/tree/main/packages/create-vite/template-lit-ts) template is doing.

Most of the plugins the package enforces are opt-in and it should be up to the user to add and configure based on their specific use-case.

rkaraivanov:
Also, sticking to ES2021 and ESM output is an acceptable default option.

The web components package targets evergreen browsers and ES2021 is a nice baseline which the _relevant_ browsers already cover.
Anything beyond that is up to user provided configuration based on their scenario.

packages/cli/templates/webcomponents/igc-ts/projects/_base/files/tsconfig.json
```diff
@@ -1,11 +1,11 @@
 {
   "compilerOptions": {
-    "target": "es2018",
-    "module": "esnext",
+    "target": "es2021",
```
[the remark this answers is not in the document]
Hristo313: @Lipata target and lib are set to ES2021 бут the module doesn't have an ES2021 option, so it's set to ES2022.

Lipata (dismissed):

packages/cli/templates/webcomponents/igc-ts/projects/_base_with_home/files/package.json
```diff
@@ -14,7 +14,7 @@
   },
   "scripts": {
     "start": "tsc && concurrently -k -r \"tsc --watch\" \"wds\"",
-    "build": "rimraf dist && tsc && node --max-old-space-size=4096 node_modules/rollup/dist/bin/rollup -c rollup.config.mjs",
+   "build": "rimraf dist && tsc && node --max-old-space-size=4096 node_modules/vite/bin/vite.js build",
```
IvayloG, line 17: Misaligned

Hristo313: The problem is reproduced in the master branch, so it's not related to this PR.

Lipata (approved):

IvayloG (approved):

PR #1365, opened 2025-01-15
chore: update dv packages with v19; standalone false for legacy components

Lipata:


valadzhov (dismissed):

valadzhov:
The stand-alone templates are fine but the legacy ones fail with the following errors:
image

Lipata:
@valadzhov you will need to add all templates for both standalone and modules, not only DV.

valadzhov (approved):
Both stand-alone and legacy projects work.

PR #1378, opened 2025-02-11
update eslint and fix warnings

Hristo313:
update packages to fix warnings for angular, react and wc when building the application.
update eslint version to 9.20


Lipata:
I've tested all the components and all scenarios for angular, react & wc.

I found one problem that is not related to lint and is already reproduced on master, that is why I logged it as a separate issue: https://github.com/IgniteUI/igniteui-cli/issues/1380.

For React there is the following problem when creating a sample with any DV component (chart, gauge, etc.) and running `build` command:

```console
[vite]: Rollup failed to resolve import "igniteui-react-core" from "C:/Work/git/igniteui-cli/output/test_react_for_reak/node_modules/igniteui-react-gauges/esm5/lib/igr-bullet-graph.js".
```
There is a related problem, when running `test` or `start`

Lipata:
The problem with react project was, because I didn't have latest node@22.14.x. Once the latter is installed all the commands on a generated project from the cli are working.

Lipata (approved):

PR #1379, opened 2025-02-24
fix test errors

Hristo313:
Fix errors when running ng/npm run test command for angular, react and wc


packages/igx-templates/igx-ts-legacy/projects/_base/files/__dot__github/workflows/node.js.yml
```diff
@@ -32,4 +32,4 @@ jobs:
     - run: npm i # replace with 'npm ci' after committing lock file from first install
     - run: npm run lint
     - run: npm run build
-    - run: npm run test -- --watch=false --browsers ChromeHeadless
+    - run: NODE_OPTIONS="--max-old-space-size=6000" npm run test -- --watch=false --browsers ChromeHeadless
```
jackofdiamond5, line 35: Not particularly sure if we should add the flag here as this will also add it for all generated project. This may not be ideal if its sole purpose is to allow the automation on generated apps to run properly. It does not seem to be needed when publishing to GitHub and running the CI - [link](https://github.com/jackofdiamond5/MY_CRM_APP/actions/runs/13499374604/job/37713794439)

Hristo313: I removed it form the yml files and added it to the package.json files

Lipata (approved):

PR #1406, opened 2025-07-01
chore: remove ng dv overrides

Lipata:


packages/igx-templates/igx-ts-legacy/projects/_base/files/package.json
```diff
@@ -30,21 +30,6 @@
     "reveal-sdk-wrappers-angular": {
```
IvayloG: The removed packages will fix the issue with **rxjs peer dep versions ** for the related components. However, in the case of **Reveal** --> We are in a Catch-22 there. Errors for both cases with OR without the override. Should be further investigated and if the override is correct as syntax config and so on, could be **npm** issue to be reported. The Reveal fix can be in a separate PR, so the majority of the clients get their fix and do not wait for the whole thing.

image

IvayloG: Meanwhile, if not too ugly, we can **override the rxjs version** to skip the client's headache of handling the error in Reveal as well, Just make sure to keep the direct dependency version in sync with the override version

image

IvayloG (approved):

PR #1497, opened 2026-02-20
fix(angular): add missing providers

### LINKED ISSUE #1496: Default Angular project is missing some dependencies setup
Repro:
- Create new project with `npx igniteui-cli@next new testProject --framework angular --type igx-ts`
- after the project install run `ig add grid` 
- Run the project and attempt to navigate `/grid`

The Grid view throws
Image
needs this to work correctly:

Image

Generated project config from base:
https://github.com/IgniteUI/igniteui-cli/blob/b8f5b53599cb20374acee53195395213060b51d3/packages/igx-templates/igx-ts/projects/_base_with_home/files/src/app/app.config.ts#L6-L9

Where the "other" base (I'll reserve harsh words for the state of project hierarchy there) comes with that and forms module setup (which might be needed too):
https://github.com/IgniteUI/igniteui-cli/blob/b8f5b53599cb20374acee53195395213060b51d3/packages/igx-templates/igx-ts/projects/_base/files/src/app/app.config.ts#L13-L18

Hristo313:
Add missing providers in the angular base and in the empty template app configs.

Closes #1496

### DIFF
--- a/packages/igx-templates/igx-ts/projects/_base_with_home/files/src/app/app.config.ts
+++ b/packages/igx-templates/igx-ts/projects/_base_with_home/files/src/app/app.config.ts
@@ -1,7 +1,7 @@
 import { ApplicationConfig, EnvironmentProviders, Provider, importProvidersFrom, provideZoneChangeDetection } from '@angular/core';
-import { provideRouter } from '@angular/router';
 import { BrowserModule, HammerModule } from '@angular/platform-browser';
 import { provideAnimations } from '@angular/platform-browser/animations';
+import { provideRouter } from '@angular/router';
 
 import { routes } from './app.routes';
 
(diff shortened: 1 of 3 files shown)


packages/igx-templates/igx-ts/projects/_base_with_home/files/src/app/app.config.ts
```diff
@@ -1,11 +1,18 @@
-import { ApplicationConfig, EnvironmentProviders, Provider, provideZoneChangeDetection } from '@angular/core';
+import { ApplicationConfig, EnvironmentProviders, Provider, importProvidersFrom, provideZoneChangeDetection } from '@angular/core';
 import { provideRouter } from '@angular/router';
```
wnvko: this should be the last import or at least auto ordering is placing it there

Issue #1575, opened 2026-03-30
CLI exposes "Base Project" option which leads to 404 error


radomirchev:
Creating new project with Ignite UI CLI leads to 404, when Base project is used. Based project must be not available option. (same for React)

Project created:

PS C:\GitHub\CLI> npm igniteui-cli
npm warn Unknown user config "email". This will stop working in the next major version of npm.
Unknown command: "igniteui-cli"

To see a list of supported npm commands, run:
  npm help
PS C:\GitHub\CLI> ig new
npm warn Unknown user config "email". This will stop working in the next major version of npm.

✔ Enter a name for your project: Hello World
✔ Choose framework: Angular
✔ Choose the type of project: Ignite UI for Angular
Psst! Did you know you can also use our schematics package with Angular CLI to create and modify your projects?
Read more at: https://www.infragistics.com/products/ignite-ui-angular/angular/components/general/cli-overview
✔ Choose project template: Base Project
✔ Choose the theme for the project: Default
  Generating project structure.
√ Project structure generated.
√ Git Initialized and Project 'Hello World' Committed

✔ Choose an action: Add component
✔ Choose a group: Grids & Lists
✔ Choose a component: Grid
✔ Choose one: Grid
✔ Name your component: Grid
√ View 'Grid' added.

✔ Choose an action: Complete & Run
The project will be created using a Trial version of Ignite UI for Angular.
You can always run the upgrade-packages command once it's created.
✔ Would you like to upgrade to the licensed feed now? Skip
✔ Choose app host port: 4200
Build started.
Installing npm packages
Packages installed successfully
Starting project.
npm warn Unknown user config "email". This will stop working in the next major version of npm.

> hello-world@0.0.0 start
> ng serve -o --port=4200

Would you like to share pseudonymous usage data about this project with the Angular Team
at Google under Google's Privacy Policy at https://policies.google.com/privacy. For more
details and how to change this setting, see https://angular.dev/cli/analytics.

   Yes

Thank you for sharing pseudonymous usage data. Should you change your mind, the following
command will disable this feature entirely:

    ng analytics disable

Global setting: enabled
Local setting: enabled
Effective status: enabled
Initial chunk files | Names         |  Raw size
styles.css          | styles        | 813.04 kB | 
main.js             | main          | 168.30 kB | 
scripts.js          | scripts       |  28.13 kB | 
polyfills.js        | polyfills     |  95 bytes | 

                    | Initial total |   1.01 MB

[...]

PR #1603, opened 2026-04-02
fix(*): replace ApplicationTitle placeholder with the real app title

### LINKED ISSUE #1579: Application title placeholder <%=ApplicationTitle%> is not replaced when creating a new app
## Description  
When creating a new application, the generated title is set to <%=ApplicationTitle%> instead of the actual application title.

## Steps to reproduce  

1.Create a new application.
2.Open the generated application.
3.Check the application title.

## Result  
The application title is rendered as <%=ApplicationTitle%>.

## Expected result  
The application title should be replaced with the correct application name provided during creation.

georgianastasov:
Closes #1579

### Description
Fixed an issue where newly generated applications could keep the literal `<%=ApplicationTitle%>` placeholder instead of using the actual application name.

### What was fixed:
- Updated the project generation flow so the application title is correctly resolved and applied.
- Covered all supported platforms: Angular, React, Web Components, and jQuery.

### Testing
- React / Web Components / jQuery: follow the standard packaged CLI flow (`build-pack` / `npm pack` / install the generated `.tgz` globally / generate a new app and verify the title).
- Angular: use the monorepo local build flow instead of the packaged CLI flow, because Angular templates come from `@igniteui/angular-templates`. Rebuild the repo, run the local CLI entrypoint, generate a new Angular app, and verify the title there.

Verified that generated apps now show the provided app name instead of the placeholder.


packages/igx-templates/igx-ts/projects/_base/index.ts
```diff
@@ -67,7 +67,7 @@ $app-palette: palette($primary, $secondary, $surface);
 			projectTemplate: this.id,
 			igxPackage: NPM_ANGULAR,
 			yamlDefaultBranch: "<%=yaml-default-branch%>", // the placeholder will be evaluated by CodeGen
-			ApplicationTitle: "<%=ApplicationTitle%>" // the placeholder will be evaluated by CodeGen
+			ApplicationTitle: name // backward-compat variable; now uses the resolved name value
```
kdinev, line 70: @georgianastasov Does codegen also need an update for this?

georgianastasov: I'm not sure actually maybe @dkalinovInfra will know better?

dkalinovInfra: I will need to check is codegen using this placeholder, but I think it will be better to not to change the placeholedr to something as generic as 'name'

damyanpetev: @dkalinovInfra @georgianastasov It's _quite_ possible this placeholder (check the history & see https://github.com/IgniteUI/igniteui-cli/pull/1237) is absolutely intentionally added like that and intentionally doesn't get replaced just for CodeGen consumption. In which case changing those at all might not be a good idea.

damyanpetev: Also possibly related to https://github.com/IgniteUI/igniteui-cli/issues/1575, but the repo steps in #1603 don't explain what type is created, assuming the base which uses those. Alt solution would be to consider hiding the base from users and clearly marking it for internal use.

Mind you, those placeholders are _not_ how this was supposed to be implemented to begin with IMO.

georgianastasov: So based on the discussion, it sounds like the `<%=ApplicationTitle%>` placeholder is intentional for CodeGen consumption and shouldn't be changed. The real fix should be hiding "Base Project" from the user-facing template list (as described in [#1575](https://github.com/IgniteUI/igniteui-cli/issues/1575), so users never encounter these pass-through placeholders. Should I close this PR and work on [#1575](https://github.com/IgniteUI/igniteui-cli/issues/1575) instead, or is there a different approach you'd prefer?

georgianastasov: @damyanpetev Yep, true.. I really did use the `base` template here, and I didn’t mention what application type gets created in the repo steps. I’ll fix the description now.

kdinev:
@georgianastasov I noticed the same behavior in React apps.

georgianastasov:
> @georgianastasov I noticed the same behavior in React apps.

Yes it is possible that it affects other platforms as well. but as we specified in the discussion above "_<%=ApplicationTitle%> placeholder is intentional for CodeGen consumption and shouldn't be changed_" so this pull request is better to be close and another fix need to be implemented. Also this happens for the `base` template which needs to be handled.. for example: "_fix should be hiding "Base Project" from the user-facing template list (as described in https://github.com/IgniteUI/igniteui-cli/issues/1575), so users never encounter these pass-through placeholders_".

PR #1698, opened 2026-05-20
feat(mcp): rewrite API URLs as get_api_reference refs with member-level lookups

dkalinovInfra:
## Description

Adds deterministic `mcp:get_api_reference?...` references for API documentation links in the MCP server's compiled docs, and extends `get_api_reference` to resolve those references at call time.

**Two coordinated changes:**

1. **Member-aware `get_api_reference`** — the tool gains an optional `member` parameter that returns only one property/method/event line instead of the full component entry. `section` (`properties` / `methods` / `events`) is now functional too — it walks the bullet list and emits only matching bullets, classified by syntax (method = signature, event = `EventEmitter<T>` or `EventCallback<T>`, otherwise property). The handler also tolerates `Component#member` shorthand defensively. Response-size impact on a real Angular grid entry: full 4.6 KB → `section=methods` 2.3 KB (−50%) → `member=data` 113 B (−97%).

2. **Deterministic URL → MCP-ref rewriter** — new `scripts/rewrite-api-links.ts` runs between `inject` and `diff` in each platform pipeline. It rewrites `https://www.infragistics.com/.../classes/<name>.html#<member>` URLs to `mcp:get_api_reference?platform=<p>&component=<C>&member=<m>`. Pure regex match per platform — no LLM in the loop. Handles all four URL shapes (bare `/docs/`, `/api/docs/` with package prefixes, Blazor FQN form, Web Components dock-manager subpath), decodes Blazor docfx synthetic anchors (`#IgniteUI_Blazor_Controls_IgbGrid_Sort` → `member=Sort`), and recovers PascalCase canonical names from lowercased URL segments via the existing `ApiDocLoader` index. Compression prompts updated to preserve `mcp:` URIs verbatim.

End-to-end smoke test against 270 Blazor doc files: **5,683 URLs rewritten across 241 files, 0 unknown-component skips, 0 mis-rewrites**.

## Related Issue

Closes #<issue-number>

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [x] Refactoring / code cleanup
- [ ] Build / CI configuration change

## Affected Packages

## Checklist

- [x] I have tested my changes locally (`npm run test`) — 173/173 passing
- [x] I have built the project successfully (`npm run build`)
- [ ] I have run the linter (`npm run lint`) — root-level lint has pre-existing parser errors in submodule files under `webcomponents/igniteui-webcomponents/**` that are unrelated to this PR; no new errors introduced by these changes
- [x] I have added/updated tests as needed — +33 new tests covering member lookup, section inference, URL rewriter regexes, and Blazor `EventCallback` classification

## Additional Context

**Sample rewrite** (`dist/docs_prepeared/blazor/grid-advanced-filtering.md`):

```diff
- [`Filterable`](https://www.infragistics.com/blazor/docs/api/api/IgniteUI.Blazor.Controls.IgbColumn.html#IgniteUI_Blazor_Controls_IgbColumn_Filterable)
+ [`Filterable`](mcp:get_api_reference?platform=blazor&component=IgbColumn&member=Filterable)
```

**Real-world `get_api_reference` probe** (`scripts/test-get-api-reference.ts`) — 27 scenarios across all four platforms, every code path passing. Response-size impact on a real `IgbGrid` entry:

| Scenario | Response size |
|---|---|
| Full entry | 6,333 B |
| `section=properties` | 1,975 B (−69%) |
| `section=methods` | 3,966 B (−37%) |
| `section=events` | 314 B (−95%) |
| `member=Data` | 45 B (−99%) |

**Section parameter was previously a no-op** — `extractSection` walked for `## Properties` / `## Methods` / `## Events` headings, but no real `llms-full.txt` file has those headings (they're flat bullet lists under `### [Component]`). The new bullet-inference fallback makes the parameter actually do what its description always claimed.

**Pipeline ordering:** `clear → (build:xplat) → export → inject → rewrite-api-urls → diff → compress → update-baseline → build:db`. Rewrite runs pre-compress so the LLM sees and preserves the final `mcp:` URIs.


packages/igniteui-mcp/igniteui-doc-mcp/scripts/rewrite-api-links.ts
```diff
@@ -0,0 +1,274 @@
...
+  regex: RegExp;
+  // Given a regex match, return the lowercase-or-PascalCase component segment.
+  // For "package.component" forms, takes the substring after the final dot.
+  extractSegment(match: RegExpExecArray): string;
+}
+
+const URL_PATTERNS: Record<Platform, UrlPattern[]> = {
+  angular: [
+    {
+      regex: /\/products\/ignite-ui-angular\/(?:api\/)?docs\/typescript\/[^/]+\/(?:classes|interfaces|enums)\/([^/)\s#]+?)\.html(?:#([^)\s]+))?/g,
```
damyanpetev: Should be noted that this can be cleaned up quite a bit in the future with native [`URLPattern`](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern) - that seems to be [Node 24+](https://nodejs.org/api/url.html#class-urlpattern) unfortunately.

```ts
const pattern = new URLPattern({
  pathname: "/products/ignite-ui-angular/:api?/docs/typescript/latest/:section(classes|enums|interfaces)/:name.html"
});
const url = "https://www.infragistics.com/products/ignite-ui-angular/docs/typescript/latest/classes/igxgridcomponent.html";
const result = pattern.exec(url)?.pathname.groups.name;
```
That'd at least make it more readable. Not sure if we can split this build script out of the CI node versions and have it with 24 only.

Also keep in mind those patterns are changing right now.

Also to note - there is/will be an ambient TS interface of that name; pretty sure that also follows the `URLPattern` naming, so it won't conflict at least.

packages/igniteui-mcp/igniteui-doc-mcp/src/tools/handlers.ts
```diff
@@ -1,12 +1,25 @@
...
+    let { component, member } = params;
+
+    // Hardening: tolerate "Component#member" being passed in `component`.
+    // The rewrite step encodes member as `&member=`, but defensive parsing here
+    // means a stray fragment-style ref still resolves.
+    if (component.includes('#')) {
+      const [head, ...rest] = component.split('#');
+      component = head.trim();
+      if (!member && rest.length > 0) {
+        const tail = rest.join('#').trim();
```
damyanpetev, line 19: Not sure why rest args are re-joined with `#` but okay. It seems like the acceptable param is a single word, so rest is unlikely to be ever be more than one, unless I'm missing something, and this is effectively `.at(0)` equivalent.

packages/igniteui-mcp/igniteui-doc-mcp/src/lib/api-doc-search.ts
```diff
@@ -55,6 +55,65 @@ export function searchApiDocs(
...
+// (the shape produced by Astro's llms-full.txt), classify each bullet from
+// its syntax:
+//   - **name**(args): T   → method (signature follows the name)
+//   - **name**: `EventEmitter<...>` → event
+//   - otherwise           → property
+function inferSectionFromBullet(line: string): MemberMatch['section'] {
+  const afterName = line.replace(/^.*?\*\*[A-Za-z_$][A-Za-z0-9_$]*\*\*/, '');
+  if (afterName.startsWith('(')) return 'method';
+  // Angular/WC/React emit events as EventEmitter<T>; Blazor uses EventCallback<T>.
+  if (/\b(?:EventEmitter|EventCallback)\b/.test(afterName)) return 'event';
```
damyanpetev: Guessing WC/React have event sections, since `EventEmitter<T>` is Angular-specific, the same way `EventCallback<T>` is ASP.NET/Blazor.

We should really consider having that handled on the API docs level somehow, as this isn't really even quite correct - pretty sure Angular also has the signal-based `output()` and that's not the same signature either. And of course, none of that is guaranteed stable, will require maintenance and really not the best to have in these scripts.

damyanpetev: Yup, it's `OutputEmitterRef<T>` - might want to include it as well. The chat component should have such but the current API docs are pending a fix to actually include that component.

kdinev (approved):

damyanpetev (approved):

PR #1701, opened 2026-05-20
fix(mcp): publish API reference docs and load latest version

dkalinovInfra:
## Description

Fixes two bugs preventing API reference data from working in the published `@igniteui/mcp-server` package, plus the CI changes needed to build those docs at release time.

**Bug 1 — `files` field shipped the wrong glob.** The `files` array in `packages/igniteui-mcp/igniteui-doc-mcp/package.json` referenced `docs/<framework>/api/...` with `.md`/`.json` extensions, but the actual on-disk layout is `docs/<framework>-api/<pkg>/<ver>/llms-full.txt`. The published tarball contained zero API entries, so `get_api_reference` / `search_api` returned "API reference not available" for every platform. Updated `files` to match the real paths and added the missing `docs/blazor-api/` entry.

**Bug 2 — `ApiDocLoader` version collision.** Entries were keyed `${platform}:${componentName}` and `readdirSync` lex-ordering meant older versions (`9.0.x`) silently overwrote newer ones (`21.0.x`) for any package with multiple version subdirs. `IgxGridComponent` was returning the v9 entry, missing fields added later (e.g. `groupingDone`). Fix has two parts:

- **Build-time filter**: `scripts/export-{angular,react,wc,blazor}-api.ts` now copy only the highest-ranked version per package and delete any older versions from the on-disk output directory. The published tarball shrinks accordingly (Angular drops `9.0.x` ~193 KB + `15.1.x` ~500 KB).
- **Runtime defence-in-depth**: `ApiDocLoader.parseLlmsFullTxt` picks the latest version when iterating, so older snapshots leaking onto disk locally cannot mask newer data.
- Both sides share a new `src/lib/version-picker.ts` so they agree on what "latest" means. Ranking: literal `latest` > parsable `<major>.<minor|x>.<patch|x>` (numeric descending, `x` outranks any concrete number) > lex fallback.

**CI changes (`.github/workflows/npm-publish.yml`).**

- Bumped `actions/checkout` and `actions/setup-node` to `@v4`.
- Added a `Build MCP API docs` step that runs `build:docs:all` so the tarball contains fresh `llms-full.txt` files at publish time.
- Configured a git global URL-rewrite using `secrets.CLASSIC_PAT_GITHUB` so the inline `git submodule update --init blazor/api-docs` can authenticate to the private `IgniteUI/api-docs` submodule.
- Added `scripts/patch-api-docs-overrides.ts` that pins `@astrojs/sitemap@^3.7.2` and `zod@^4.3.6` in the submodule's `package.json` before its `npm install` — works around an upstream Astro `z.function(...).optional is not a function` runtime error.

## Related Issue

Closes #<issue-number>

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring / code cleanup
- [x] Build / CI configuration change

## Affected Packages

## Checklist

- [ ] I have tested my changes locally (`npm run test`)
- [ ] I have built the project successfully (`npm run build`)
- [ ] I have run the linter (`npm run lint`)
- [x] I have added/updated tests as needed

## Additional Context

- New test files: `src/__tests__/lib/version-picker.test.ts` (12 cases — semver ordering, `latest` precedence, `x` in minor and patch slots, empty-input throw, input-not-mutated) and a regression test in `src/__tests__/lib/api-doc-loader.test.ts` for the multi-version scenario.
- Verification: `npm pack --dry-run` in `packages/igniteui-mcp/igniteui-doc-mcp/` now lists `docs/*-api/**/llms-full.txt` entries (was zero before the `files` fix).
- A `CLASSIC_PAT_GITHUB` repo secret with `repo` scope on the `IgniteUI` org is required for the workflow to clone the private `IgniteUI/api-docs` submodule. Token expiration will cause future releases to fail until rotated — long term, consider migrating to a GitHub App so this isn't a recurring concern.


.github/workflows/npm-publish.yml
```diff
@@ -12,8 +12,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.CLASSIC_PAT_GITHUB }}
```
damyanpetev, line 17: If you're adding a PAT `token` to this step as well, can we not just use `submodules: true/recursive` and have those clones correctly? I assume all of the submodules are needed at build time.

That'll also allow dropping or at least splitting the spread out submodule update calls from build scripts.

kdinev (dismissed):

packages/igniteui-mcp/igniteui-doc-mcp/scripts/patch-api-docs-overrides.ts
```diff
@@ -0,0 +1,42 @@
+import { existsSync, readFileSync, writeFileSync } from 'fs';
+import { join, resolve } from 'path';
+
+const ROOT = resolve(import.meta.dirname, '..');
+const PKG_PATH = join(ROOT, 'blazor', 'api-docs', 'package.json');
+
+const OVERRIDES = {
+  '@astrojs/sitemap': '^3.7.2',
```
damyanpetev: Um any specific reason for the overrides? Kinda doesn't seem it should be needed, the submodules should be hitting stable refs, right?
Besides the fact this can affect the lock file, it will also go out of sync with the submodule repo - so fragile either way.

If there's an issue, is this something that should be addressed/fixed in the submodule repo itself?

packages/igniteui-mcp/igniteui-doc-mcp/src/lib/version-picker.ts
damyanpetev, line 1: And this one is a bit shaky too, though almost all of the scripts around collecting those API docs are prone to breakage if anything shifts slightly in the API docs build process.

Unless speed is critical here, as this is a requirement for the process, it should be possible to address this in the submodule repo again - either by producing artifacts that can be consumed by the MCP eventually, but even before that on the build commands level in api-docs we should really have a sub-set command to produce _just_ the latest API Docs for a platform _and_ even just the LLMs txt if possible and even preferably with parametrized path so this process can call something like:
```bash
npm run build:angular-latest-llm:en -- ../../docs/angular-api/llms-full.txt
```
and be pretty much done with the whole process.

packages/igniteui-mcp/igniteui-doc-mcp/scripts/patch-api-docs-overrides.ts
damyanpetev: PS: Really odd the api-docs for all is cloned under blazor 😄

.github/workflows/npm-publish.yml
dkalinovInfra: That PAT token was already added, so I've decided to reuse it. About other submodules, for now we don't use them during the build, so this will slow down the release workflow without any real need to have all of the submodules.

packages/igniteui-mcp/igniteui-doc-mcp/scripts/patch-api-docs-overrides.ts
dkalinovInfra: I will probably deleted this script, as it may not be needed any more. Api repo was updated with same sitemap version as we need for the build (the issue was coming from zod versions sitemap needs v3 but we have dependency on v4 and this was breaking the build).

damyanpetev (dismissed):

PR #1702, opened 2026-05-21
fix(igx-templates): fix side-nav layout stacking and scrolling issues

### LINKED ISSUE #1615: [CLI MCP] [Angular]: Generated app layout stacks nav drawer and routed content vertically
### Description
In the generated Angular app, the navigation drawer and the main routed content are rendered as sibling elements, but the root app container does not define a layout that places them side by side.

Because of this, the browser uses the default document flow and stacks them vertically instead:

- the drawer appears first
- the main page content is rendered below it

This makes the app shell look broken and does not match the expected layout.

### Proposed Solution
The fix is in styles: the root app container should use a two-column layout so the drawer stays on the left and the main content stays on the right. The main content area should also include min-width: 0 and min-height: 0 so pages shrink and scroll correctly without overflow issues.

ivanvpetrov:
Closes #1615

## Description

Fixes the Angular `side-nav` and `side-nav-auth` generated app shell layout where:
- The navigation drawer and main content were stacked vertically instead of side by side
- The navigation drawer would scroll out of view when the page content was taller than the viewport

## Root Cause

The `.outer-wrapper` container relied solely on the `igxLayout` directive for `display: flex`, but had no explicit `overflow: hidden` to constrain the flex container to its height. Because `html/body` already have `height: 100%` but `overflow: visible`, page-level scrolling was triggered for tall content — causing the nav drawer to scroll away with the page instead of staying fixed.

Additionally, `side-nav-auth` was missing its `app.scss` file entirely, even though `app.ts` referenced it via `styleUrl: './app.scss'`.

## Changes

- **`side-nav/files/src/app/app.scss`**: Added `display: flex; flex-direction: row` to `.outer-wrapper`, added `overflow: hidden` to `app-root`, `.outer-wrapper`, and `.main`, and added `.main` rule with `flex: 1; min-width: 0; min-height: 0`
- **`side-nav/files/src/app/app.html`**: Added `class="main"` to the inner content `<div>` so it can be targeted by the `.main` CSS rule
- **`side-nav-auth/files/src/app/app.scss`**: Created the missing stylesheet with the same layout rules (the `side-nav-auth` template already had `class="main"` on the inner div)

## Result

The nav drawer and navbar are now stationary. Only the routed page content area scrolls via `overflow: auto` on `.content`.

## Preview before fix

https://github.com/user-attachments/assets/ef03c46f-37da-4d76-9d39-9a349c1117a8

## Preview after fix

https://github.com/user-attachments/assets/4ad60817-65de-46c1-8198-3779b4015401


packages/igx-templates/igx-ts/projects/side-nav/files/src/app/app.html
```diff
@@ -7,7 +7,7 @@
       }
     </ng-template>
   </igx-nav-drawer>
-  <div igxFlex igxLayout igxLayoutDir="columns">
+  <div class="main" igxFlex igxLayout igxLayoutDir="columns">
```
damyanpetev, line 10: @ivanvpetrov Can you check the directives are actually active?
`igxLayout` on the `.outer-wrapper` _is_ supposed to place it in `display: flex;` IIRC.
Simlartly with `igxFlex` on the `.main`. I'm assuming it's just the imports are missing?

Might want to check out https://www.infragistics.com/products/ignite-ui-angular/angular/components/layout to see what those directives added, recreate that in CSS as a starting point and actually remove the directives if they won't be used anymore.

damyanpetev: Also not sure what scrolling issues there are, but you shouldn't have to spam that much `overflow: hidden;` (if at all) in a cleanly set up layout.

ivanvpetrov:
This behavior was occurring most likely because of the LLM trying to generate (and sometimes failing) a side nav + mini side nav layout, which was not available out of the box in the CLI. 
Such templates are now incorporated in PR #1716 and would be pick up by the LLMs

